Issue #149, opened 2026-04-24
3.13 — Support array of event types in event filter


determined-001:
**Complexity:** Medium — 150 Points · `type:feature`

**Context**
`event?: string` accepts a single type or `"*"`.

**Problem**
A consumer that wants both `payment.received` and `payment.sent` (but not `engine.*` notifications) has to subscribe twice.

**What needs to be done**
1. Widen `event` to `string | string[]`.
2. Server-side filtering (in `useEffect`) accepts the union.
3. README example.

**Key files**
- `packages/pulse-notify/src/index.ts:7`, `:46`.

**Done when**
- An array of event types narrows the subscription correctly.

martoniel:
@martoniel has applied to work on this issue as part of the **Stellar Wave Program**'s 4th wave.


> I’d like to work on this issue. I can extend the event parameter to support string | string[], allowing consumers to subscribe to multiple specific events like payment.received and payment.sent in a single hook.
> 
> I’ll update the server-side filtering logic in useEffect to handle both types and add a clear README example demonstrating usage. 
> 
> I can solve this issue within 24 hours, just like I’ve done with other issues I’ve been assigned to.
> 
> 


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/a33595ac-64f5-40c3-bb62-ea46583f0fcd) or assign @martoniel to this issue.

PR #213, opened 2026-04-24
Closes 3.1: UseEffect dep Array

### LINKED ISSUE #137: 3.1 — useEffect dep array should list individual config fields
**Complexity:** Trivial — 100 Points · `type:bug` · `good-first-issue`

**Context**
`useStellarEvent` depends on `[config.serverUrl, config.address, config.event, config.token]` — which is correct. But the README and convenience hooks construct a fresh `config` object every render.

**Problem**
The dep array shape is fine, but consumers passing a fresh `{...}` literal each render wouldn't be helped by changing it to depend on the object identity. Confirm the existing dep array, and add a `useMemo` example to the README — or better, support passing primitives directly.

**What needs to be done**
1. Verify the `useEffect` dep array references each primitive field, not the object.
2. Add a "Stable config" section to README warning against inline object literals.
3. Optionally provide a primitives-first overload `useStellarEvent(serverUrl, address, options)`.

**Key files**
- `packages/pulse-notify/src/index.ts:30`, `:66`.
- `packages/pulse-notify/README.md`.

**Done when**
- Dep array is verified correct and the README warns about the object-identity pitfall.

martoniel:
Closes #137

---

## Linked issue

Closes #

## Test plan

## Breaking changes

None / Yes (describe below)

### DIFF
--- a/packages/pulse-notify/README.md
+++ b/packages/pulse-notify/README.md
@@ -40,26 +40,30 @@ export function LiveBalance({ address }: { address: string }) {
 
 ## Hooks
 
-### `useStellarEvent(config)`
+### `useStellarEvent(serverUrl, address, options?)`
 
 The core hook. Subscribes to any event type on an address.
 
 ```ts
-const { event, connected, error } = useStellarEvent({
-  serverUrl: "https://events.example.com",
-  address: "GABC...",
-  event: "payment.received", // or "*" for all
-  token: "orb_live_abc...",  // optional API key
-});
+const { event, connected, error } = useStellarEvent(
+  "https://events.example.com",
+  "GABC...",
+  {
+    event: "payment.received", // or "*" for all (default)
+    token: "orb_live_abc...",  // optional API key
+  }
+);
 ```
+
+Also accepts a single config object — see [Stable config](#stable-config) for when that matters.
 
 ### `useStellarPayment(serverUrl, address)`
 
-Shorthand for payments received. Equivalent to `useStellarEvent({ ..., event: "payment.received" })`.
+Shorthand for payments received. Equivalent to `useStellarEvent(serverUrl, address, { event: "payment.received" })`.
 
 ### `useStellarActivity(serverUrl, address)`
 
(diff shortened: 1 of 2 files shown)


determined-001 (approved):
## Review: Approve — but coordinate with #214 before merging

The dep array fix itself is correct. The old `[config.serverUrl, config.address, config.event, config.token]` was relying on `config`'s primitive fields, which works, but it left the `config` object identity in the dep array — a trap for callers who construct it inline. Extracting the four primitives first, then depending on those directly, is the right fix. The overloaded signature is also clean.

### Conflict with #214

This PR and #214 (`fix/Support-array-of-event`, also by you) make identical changes to the same lines of `packages/pulse-notify/src/index.ts` — both add the overloaded signature and fix the dep array to use extracted primitives. #214 is a superset: it includes all of this PR's changes **plus** `event?: string | string[]` array support.

Two options:

**Option A — Merge this first, then #214 rebases**: Merge #213 (the focused dep-array fix), then update #214 to remove the duplicate overload/dep-array changes, keeping only the array-event additions. This keeps each PR scoped to one concern and makes the git history readable.

**Option B — Close this in favour of #214**: If you want both concerns in one shot, close #213 with a note that #214 covers it, and let #214 claim both issue closes (`Closes #137, Closes #149`).

Either is fine. Just avoid merging both as-is — they'll conflict and one will introduce a duplicate implementation.

### Other issues

**PR description is completely blank** — no "What changed and why", no test plan checked. Please fill it in before merging.

**No tests added** — the test checkbox is unchecked and no test file exists for `pulse-notify`. The dep array serialisation and the overload normalisation are worth unit testing (e.g., check that `useStellarEvent(url, addr, opts)` and `useStellarEvent({ serverUrl: url, address: addr, ...opts })` produce identical dep-array values).

determined-001:
Closing as superseded by #214. All changes in this PR — the overloaded call signature and the dep-array primitive extraction — were included in #214, which also added array event support. Both fixes are now on main. Thanks for the contribution, @martoniel!

PR #214, opened 2026-04-24
Closes 3.13: Support Array of Events

### LINKED ISSUE #149: 3.13 — Support array of event types in event filter

martoniel:
Closes #149

---

## Linked issue

Closes #

## Test plan

- [x] Manually tested against testnet / mainnet (describe what you tested)

## Breaking changes

None / Yes (describe below)

### DIFF
--- a/packages/pulse-notify/README.md
+++ b/packages/pulse-notify/README.md
@@ -40,26 +40,42 @@ export function LiveBalance({ address }: { address: string }) {
 
 ## Hooks
 
-### `useStellarEvent(config)`
+### `useStellarEvent(serverUrl, address, options?)`
 
-The core hook. Subscribes to any event type on an address.
+The core hook. Subscribes to one event type, an allowlist of types, or all events on an address.
 
 ```ts
-const { event, connected, error } = useStellarEvent({
-  serverUrl: "https://events.example.com",
-  address: "GABC...",
-  event: "payment.received", // or "*" for all
-  token: "orb_live_abc...",  // optional API key
-});
+// Single type
+const { event, connected, error } = useStellarEvent(
+  "https://events.example.com",
+  "GABC...",
+  { event: "payment.received" }
+);
+
+// Multiple types — subscribe to both without opening two connections
+const { event } = useStellarEvent(
+  "https://events.example.com",
+  "GABC...",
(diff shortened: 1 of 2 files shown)


determined-001 (approved):
## Review: Approve with Comments

The implementation is solid. The overloaded signature pattern is clean, and serialising the event array to a sorted string for the dep array is the correct way to prevent reference-equality churn without exposing the problem to callers. The README additions are thorough.

### Issues to address

**1. Test checkbox is checked but no tests were added**
The diff only touches `README.md` and `src/index.ts`. The PR template has "New tests added for new behaviour" checked, but there are no test files anywhere in `packages/pulse-notify/` (not in this PR, and none existed before). The array filter path — especially the sorting logic and the effect stability guarantee — is the kind of behaviour that should be covered.

At a minimum: a test that verifies the `eventKey` serialisation is order-independent (i.e. `["payment.sent", "payment.received"]` and `["payment.received", "payment.sent"]` produce the same stable key) and a test that the filter correctly allows/blocks incoming events.

**2. PR description was not filled out**
The "What changed and why" and "Notes for reviewers" sections are still the template placeholders. Please fill them in — it helps future contributors understand the motivation for the change when reading git history.

### Minor notes

- The `address!` non-null assertion in the implementation overload is safe because the call-signature overloads enforce `address: string`, but a comment or a runtime guard would make the intent clearer to future editors.
- The dep-array comment (`// ✅ eventKey is a serialised string…`) is good — keep it. This is one of the cases where the why is genuinely non-obvious.
- Sorting before joining is the right call; `["a","b"]` and `["b","a"]` should be treated as the same subscription.

Overall: the logic is correct and the docs are well-written. Just needs a test file and a filled-out description before merging.

PR #215, opened 2026-04-24
normalize set_options into account.options_changed event

### LINKED ISSUE #81: 1.21 — Normalize set_options

**Context**
`set_options` covers signer changes, threshold changes, home domain changes, and inflation destination changes.

**Problem**
Multisig and account-config tooling can't observe these changes via Pulse.

**What needs to be done**
1. Normalize `set_options` into a single `type: "account.options_changed"` event.
2. Diff the present fields and include them in the payload (`signer_added`, `signer_removed`, `thresholds`, `home_domain`, etc.).
3. Route to the source-account watcher.
4. Unit tests covering signer add, signer remove, threshold change, home domain change.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:207`, `:230`.

**Done when**
- `set_options` emits a single event whose payload only includes fields the operation actually changed.

blockchainrafik:
## Linked issue

Closes #81

- Switch operations stream from payments() to operations() so set_options
  records are received
- Add normalizeSetOptions() that diffs signer_key/weight, low/med/high
  thresholds, master_key_weight, and home_domain; only changed fields are
  included in the payload
- Route account.options_changed to the source-account watcher
- Add AccountOptionsEvent, AccountOptionsChanges, SetOptionsSigner types;
  widen NormalizedEvent to PaymentEvent | AccountOptionsEvent
- Widen Watcher on/emit to accept WatcherNotification alongside NormalizedEvent
- 7 new unit tests: signer_added, signer_removed, threshold change,
  home_domain change, field exclusivity, no-op suppression, routing isolation

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -1,18 +1,21 @@
 import { Horizon } from "@stellar/stellar-sdk";
 import { Watcher } from "./Watcher.js";
 import type {
+  AccountOptionsChanges,
+  AccountOptionsEvent,
+  AccountOptionsEventType,
   CoreConfig,
   Network,
   NormalizedEvent,
+  PaymentEvent,
   PaymentEventType,
   ReconnectConfig,
   WatcherNotification,
   WatcherNotificationType,
 } from "./index.js";
 
-type PendingNormalizedEvent = Omit<NormalizedEvent, "type"> & {
-  type: "unknown";
-};
+type PendingPaymentEvent = Omit<PaymentEvent, "type"> & { type: "unknown" };
+type PendingNormalizedEvent = PendingPaymentEvent | AccountOptionsEvent;
 
 type StreamCallbacks = {
   onmessage: (record: unknown) => void;
@@ -131,7 +134,7 @@ export class EventEngine {
     };
 
     this.stopStream = this.server
-      .payments()
+      .operations()
       .cursor("now")
       .stream(callbacks) as StreamStopper;
   }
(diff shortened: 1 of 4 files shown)


determined-001 (approved):
## Review

### Issues

**1. `signer_weight` cast without a null guard — medium**

```ts
if (typeof r.signer_key === "string") {
  const weight = r.signer_weight as number;  // unsafe cast
  if (weight === 0) { ... } else { ... }
}
```

If Horizon ever returns `signer_key` without `signer_weight` (malformed or future spec change), `weight` is `undefined as number`. `undefined === 0` is `false`, so it falls into `signer_added` with `weight: undefined` silently. Safer:

```ts
const weight = typeof r.signer_weight === "number" ? r.signer_weight : 0;
```

**2. `set_flags`, `clear_flags`, and `inflation_dest` are silently swallowed — minor**

An op that only changes flags (e.g. enabling `AUTH_REQUIRED`) produces no `changes`, hits the `Object.keys(changes).length === 0` guard, and is silently dropped as a no-op. This is probably intentional scope, but should either be tracked in a follow-up issue (reference it in a comment at the `set_options` guard) or noted in the PR description so it reads as a known gap, not an oversight.

**3. `PendingNormalizedEvent` naming is misleading — minor**

```ts
type PendingNormalizedEvent = PendingPaymentEvent | AccountOptionsEvent;
```

`AccountOptionsEvent` is not "pending" — its type is fully resolved as `"account.options_changed"`. Naming it `PendingNormalizedEvent` implies it needs further resolution. A clearer name would be `NormalizedEventOrPending` or just an inline union at the one call site.

**4. Public type rename is a breaking change — worth calling out**

The old `NormalizedEvent` (with `.to`, `.from`, `.amount`, `.asset`) is now `PaymentEvent`. Any downstream consumer importing `NormalizedEvent` and accessing those fields will get a TypeScript error after this merges. Correct at `v0.x`, but the PR description and commit message don't flag it. Worth noting so the changelog captures it.

---

### What is good

- Switching to `.operations()` is the right call — `.payments()` would never surface `set_options`.
- The diff approach in `normalizeSetOptions` (only emit fields that actually changed) is exactly right.
- No-op suppression (`Object.keys(changes).length === 0 → return null`) is correct and tested.
- `signer_removed` detection via `weight === 0` matches the Stellar spec.
- Routing by `source_account` is correct — `set_options` has no `to`/`from`.
- `withResolvedType` narrowed to `PendingPaymentEvent` is a clean improvement.
- 7 tests cover the important cases including routing isolation. `makeSetOptionsRecord` keeps them readable.

The `signer_weight` null guard (#1) is worth fixing before merge — it is a one-liner that avoids a silent bad-data path. The rest are non-blocking.

blockchainrafik:
fixed nullguard

PR #217, opened 2026-04-24
fix: set 16kb body size limit on express.json()

### LINKED ISSUE #162: 4.2 — Request body size limits
**Complexity:** Trivial — 100 Points · `type:security` · `good-first-issue`

**Context**
`express.json()` defaults to a 100kb limit, but the limit isn't documented and an attacker who finds the default can target it.

**Problem**
Webhook registration bodies should be small (< 4kb). The default 100kb is generous.

**What needs to be done**
1. Pass `{ limit: "16kb" }` to `express.json()`.
2. Document the limit in the server README.

**Key files**
- `apps/server/src/index.ts:39`.

**Done when**
- Body limit is explicit and documented.

Clement-coder:
- Pass { limit: '16kb' } to express.json() in apps/server/src/index.ts
- Document the limit in apps/server/README.md security defaults

## Linked issue

Closes #162

## What changed and why

Sets an explicit 16kb body size limit on express.json() in the reference server.
The Express default of 100kb is unnecessarily generous for webhook registration
payloads, which are small by design. Making the limit explicit removes 
ambiguity and reduces the attack surface for oversized-body abuse. The limit is 
also documented in the server README under Security defaults.

## Test plan

- [ ] Existing tests pass (pnpm test)
- [ ] Typechecks pass (pnpm -r typecheck)

## Breaking changes

None

## Notes for reviewers

The only functional change is express.json() → express.json({ limit: "16kb" }). 
Requests with a body larger than 16kb will now receive a 413 Payload Too Large 
response. No existing endpoint accepts bodies anywhere near that size, so no 
legitimate client should be affected.

### DIFF
--- a/apps/server/README.md
+++ b/apps/server/README.md
@@ -66,6 +66,7 @@ useStellarEvent({
 
 ## Security defaults
 
+- **Request body limit.** `express.json()` enforces a **16 kb** maximum body size. Webhook registration payloads are small by design; oversized requests are rejected with `413 Payload Too Large`.
 - **HTTPS enforcement.** The server rejects `http://`, `localhost`, and private-IP-range webhook URLs at registration time to prevent SSRF.
 - **Stellar key validation.** Registered addresses must pass `StrKey.isValidEd25519PublicKey` before they're accepted.
 - **Secrets are hashed.** Webhook secrets are HMAC-hashed before storage; the plaintext never persists beyond the request.
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -36,7 +36,7 @@ console.log(`[server] Event engine started on ${NETWORK}`);
 const registry = new WebhookRegistry(engine);
 
 const app = express();
-app.use(express.json());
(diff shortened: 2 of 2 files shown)


determined-001 (approved):
## Review: Approve

One-line change, correct rationale, and the PR description is the best-written one in this batch — it explains exactly what changed, why, and what the observable impact is.

### Notes

- `"16kb"` is the correct format for Express body-parser's `limit` option. No issues there.
- The server has no global error-handler middleware, so a 413 will be handled by Express's built-in default. That produces a plain-text `PayloadTooLargeError` response — acceptable for a reference server. If this ever needs a JSON error body, add a global `(err, req, res, next)` handler, but that's out of scope here.
- None of the test-plan boxes are checked, which is honest — this is hard to cover without a running server. A note in the PR description about what you'd need to test manually (e.g., `curl -X POST -H 'Content-Type: application/json' -d '<body > 16kb>'`) would help future maintainers verify the behaviour.

Clean change. Ready to merge.

PR #221, opened 2026-04-26
Feat/fix pulse notify

### LINKED ISSUE #152: 3.16 — TypeScript generics for app-level event narrowing
**Complexity:** Medium — 150 Points · `type:dx`

**Context**
After #1.22, `NormalizedEvent` becomes a discriminated union. Consumers want to constrain the union to their app's expected events.

**Problem**
The hook's `event: NormalizedEvent | null` is too wide for many consumers.

**What needs to be done**
1. Make the hook generic: `useStellarEvent<T extends NormalizedEvent = NormalizedEvent>(config)` returning `EventState<T>`.
2. Type-narrow inside the hook based on the `event` filter where possible.
3. README example showing app-level type narrowing.

**Key files**
- `packages/pulse-notify/src/index.ts:23`.

**Done when**
- A consumer can narrow the event type via the generic and get full IDE support.

chizzy192:
## Linked issue

Closes #152 

## Test plan

- [x] Manually tested against testnet / mainnet (describe what you tested)

## Breaking changes

None / Yes (describe below)

### DIFF
--- a/packages/pulse-notify/README.md
+++ b/packages/pulse-notify/README.md
@@ -80,12 +80,48 @@ Shorthand for all events on an address. Equivalent to `useStellarEvent(serverUrl
 ## Return shape
 
 ```ts
-type EventState = {
-  event: NormalizedEvent | null; // Latest event, or null before first arrival
-  connected: boolean;            // True once the SSE handshake completes
-  error: string | null;          // Error message if the connection fails
+type EventState<T extends NormalizedEvent = NormalizedEvent> = {
+  event: T | null;     // Latest event, or null before first arrival
+  connected: boolean;  // True once the SSE handshake completes
+  error: string | null; // Error message if the connection fails
 };
 ```
+
+## Type narrowing
+
+`useStellarEvent` is generic — pass a narrower union as `T` to get full IDE support and avoid manual casts. Use TypeScript's `Extract` to pull specific event types out of `NormalizedEvent`:
+
+```tsx
+import type { NormalizedEvent } from "@orbital/pulse-core";
+import { useStellarEvent } from "@orbital/pulse-notify";
+
+type WalletEvents = Extract<
+  NormalizedEvent,
+  { type: "payment.received" | "payment.sent" | "trustline.added" }
+>;
+
+function Wallet({ address }: { address: string }) {
+  const { event } = useStellarEvent<WalletEvents>(
+    "https://events.example.com",
+    address,
+    { event: ["payment.received", "payment.sent", "trustline.added"] }
+  );
+
+  if (!event) return null;
+
+  // event.type is now "payment.received" | "payment.sent" | "trustline.added"
+  // — TS narrows the rest of the shape per branch.
+  switch (event.type) {
+    case "payment.received":
+    case "payment.sent":
+      return <div>{event.amount} {event.asset}</div>;
+    case "trustline.added":
+      return <div>Added {event.asset}</div>;
+  }
+}
+```
+
+The default `T = NormalizedEvent` keeps the existing untyped behavior — pass `<T>` only when you want narrowing.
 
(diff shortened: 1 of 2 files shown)


determined-001 (changes requested):
## Review: Request Changes

### Issue 1 — Array event filtering is removed (regression against #214)

PR #214 added array event type support:
```ts
// Before this PR:
const allowed =
  eventType === "*" ||
  (Array.isArray(eventType)
    ? eventType.includes(incoming.type)
    : incoming.type === eventType);
```

This PR replaces it with:
```ts
const matchesConfiguredType =
  eventType === "*" || incoming.type === eventType;
```

`incoming.type === eventType` always returns `false` when `eventType` is an array, so `useStellarEvent(url, addr, { event: ["payment.received", "payment.sent"] })` now receives nothing. This silently breaks every caller using an event array.

### Issue 2 — `useStellarEvent` overload is removed (regression against #214)

The primitives-first overload `useStellarEvent(serverUrl, address, options?)` is gone. Callers using that signature will get a TS error after merging.

### Issue 3 — Generic `T` doesn't filter at runtime, only at compile time

```ts
setState((prev) => ({ ...prev, event: incoming as T }));
```

The cast `as T` only satisfies the TypeScript compiler. If `T = PaymentEvent` and the SSE stream sends an `AccountOptionsEvent`, the hook stores a wrongly-typed value. Consumers comparing `event.type` at runtime won't be affected, but any code relying on `T`-specific fields would get `undefined` silently.

This is acceptable if documented ("T narrows the TypeScript type only — filter by `event.type` at runtime"), but the current JSDoc example doesn't warn about it.

### Issue 4 — `pnpm-lock.yaml` downgrades packages already on main

The lockfile rolls back typescript 6.0.3 → 6.0.2, postcss 8.5.10 → 8.5.8, `@types/node` 25.6.0 → 25.5.0, vitest 4.1.5 → 4.1.2 etc. — all versions that were already bumped to current on main by the merged Dependabot PRs. Please rebase on main and run `pnpm install` to pick up the current lockfile.

### Minor notes

- `tsc -b` with `composite: true` and project references is a valid improvement for incremental builds, no issues there.
- Renaming `WatcherEvent` → `WatcherEventPayload` and exporting it is fine, but does expand the public API surface.

determined-001 (commented):
Thanks @chizzy192! After a careful look, the branch is significantly out of sync with current `main` — `useStellarEvent` was refactored into a TypeScript overload pattern (config-form OR positional-form, single implementation), so applying this PR as-is would lose that refactor and ship a few unintended regressions. Since you're not available, I'm taking option 2 here: rewinding the branch to current `main` and applying just the parts of the change that match #152.

**What I'm keeping (matches #152):**
1. `EventState` becomes generic: `EventState<T extends NormalizedEvent = NormalizedEvent>` with `event: T | null`.
2. Both `useStellarEvent` overloads accept the generic `<T>` and return `EventState<T>`.
3. `useStellarPayment` narrows via `Extract<NormalizedEvent, { type: "payment.received" }>` so consumers get a typed `PaymentEvent` directly.
4. README gets a "Type narrowing" section with the example.

**What I'm dropping (out of scope for #152, would regress current behavior):**
- Removing the `(serverUrl, address, options)` overload — current main has it as a deliberate refactor.
- Dropping `string[]` array event filter — current main supports allowlists; removal would silently drop events.
- `tsc -p` → `tsc -b` + `composite: true` + project references on `pulse-notify` — out of scope and affects the workspace build pipeline.
- Renaming/exporting `WatcherEvent` → `WatcherEventPayload` from `pulse-core` — unrelated to #152.

Heads up: the branch will be force-pushed because main has moved past where this was based. Your authorship will still be on the squash-merge as the PR author. Thanks for the original work — it surfaced the gap in #152's typing.

PR #222, opened 2026-04-26
fix: drop redundant registry.clear() in EventEngine.stop() (#64)

### LINKED ISSUE #64: 1.4 — Drop redundant registry.clear() in EventEngine.stop()
**Complexity:** Trivial — 100 Points · `type:refactor` · `good-first-issue`

**Context**
`EventEngine.stop()` iterates each watcher and calls `watcher.stop()`, then calls `this.registry.clear()`.

**Problem**
Each watcher's stop handler already removes its entry from the registry (`EventEngine.subscribe` registers exactly that handler at line 60). The `clear()` call is dead code that obscures the lifecycle contract.

**What needs to be done**
1. Remove the trailing `this.registry.clear()` in `EventEngine.stop()`.
2. Add a unit test asserting `registry.size === 0` after `stop()` so the invariant is locked in.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:93` — line to remove.
- `packages/pulse-core/src/EventEngine.ts:60` — the stop handler that already does the work.

**Done when**
- `stop()` no longer calls `registry.clear()`.
- A test confirms the registry is emptied via stop handlers alone.

Awointa:
What changed

Removed the trailing this.registry.clear() call from EventEngine.stop().

Why

Each watcher's stop handler already deletes its own entry from the registry — this is registered in subscribe() at line 60. The clear() call was dead code that made the lifecycle contract harder to reason about, implying the registry needed a manual sweep when it didn't.

Test

Added a unit test that subscribes two watchers, calls stop(), and asserts registry.size === 0 — confirming the invariant is upheld by the stop handlers alone, without any explicit clear.

Closes #64

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -94,8 +94,6 @@ export class EventEngine {
     for (const watcher of this.registry.values()) {
       watcher.stop();
     }
-
-    this.registry.clear();
   }
 
   private openStream(isReconnect: boolean): void {
--- a/packages/pulse-core/test/pulse-core.test.ts
+++ b/packages/pulse-core/test/pulse-core.test.ts
@@ -102,6 +102,20 @@ describe("pulse-core EventEngine", () => {
     });
   });
 
+  it("empties the registry via stop handlers when stop() is called", () => {
+    const engine = new EventEngine({ network: "testnet" });
+    engine.subscribe("GABC");
+    engine.subscribe("GDEF");
+
+    const registry = (engine as unknown as { registry: Map<string, unknown> })
+      .registry;
+    expect(registry.size).toBe(2);
+
+    engine.stop();
+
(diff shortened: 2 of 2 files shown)


determined-001 (approved):
## Review: Approve

`registry.clear()` after `for (watcher of registry.values()) watcher.stop()` is indeed unreachable dead code. Each `watcher.stop()` triggers `onStop` → `registry.delete(address)`, so the registry is empty by the time `registry.clear()` runs. The test proves it cleanly — subscribing two watchers, stopping the engine, then asserting `registry.size === 0` is exactly the right verification.

One small note: the test accesses `(engine as unknown as { registry: Map<string, unknown> }).registry` to reach the private field. This is fine for a unit test, but if the existing test suite already has a pattern for accessing private engine state, keep it consistent.

determined-001:
Thank you for your contribution

PR #224, opened 2026-04-26
feat: multiple-delivery-ur-ls-per-watcher-fan-out

### LINKED ISSUE #121: 2.15 — Multiple delivery URLs per watcher (fan-out)

**Context**
Today a `WebhookDelivery` has exactly one URL.

**Problem**
A single subscription needs multiple downstream consumers (production receiver + staging mirror + internal audit log) without subscribing the same address three times.

**What needs to be done**
1. Allow `WebhookConfig.url` to be `string | string[]`.
2. On every event, deliver to each URL in parallel; track retries independently per URL.
3. Each per-URL failure emits its own `webhook.failed` keyed on the URL.

**Key files**
- `packages/pulse-webhooks/src/index.ts:8`, `:36`.
- `packages/pulse-webhooks/src/types.ts:1`.

**Done when**
- A single delivery config can fan out to N URLs with independent retry state.

3m1n3nc3:
## Linked issue

Closes #121 

## What changed and why

This PR adds fan-out delivery support to @orbital/pulse-webhooks so a single watcher can deliver the same event to multiple downstream webhook URLs.

## Test plan

## Breaking changes

None

### DIFF
--- a/apps/web/content/api/pulse-webhooks.md
+++ b/apps/web/content/api/pulse-webhooks.md
@@ -5,7 +5,7 @@ description: HMAC-signed webhook delivery with automatic retry.
 
 ## Overview
 
-`@orbital/pulse-webhooks` wraps a `Watcher` and delivers events to an HTTP endpoint. Each delivery is signed with HMAC-SHA256 so your server can verify authenticity.
+`@orbital/pulse-webhooks` wraps a `Watcher` and delivers events to one or more HTTP endpoints. Each delivery is signed with HMAC-SHA256 so your server can verify authenticity.
 
 ## Installation
 
@@ -21,7 +21,10 @@ npm install @orbital/pulse-webhooks
 import { WebhookDelivery } from '@orbital/pulse-webhooks'
 
 const delivery = new WebhookDelivery(watcher, {
-  url: 'https://your-app.com/webhook',
+  url: [
+    'https://your-app.com/webhook',
+    'https://staging.your-app.com/webhook',
+  ],
   secret: 'my-signing-secret',
   retries: 3,               // optional, default: 3
   deliveryTimeoutMs: 10000, // optional, default: 10000
@@ -32,7 +35,7 @@ const delivery = new WebhookDelivery(watcher, {
 
 | Option | Type | Default | Description |
 |--------|------|---------|-------------|
-| `url` | `string` | — | Endpoint to POST events to |
+| `url` | `string \| string[]` | — | Endpoint to POST events to, or a list of endpoints for fan-out |
 | `secret` | `string` | — | HMAC signing secret |
 | `retries` | `number` | `3` | Max delivery attempts |
 | `deliveryTimeoutMs` | `number` | `10000` | Per-attempt timeout in ms |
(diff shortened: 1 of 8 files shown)


determined-001 (approved):
## Review: Approve with one request

The fan-out implementation is clean. `ResolvedWebhookConfig` normalising `url` to `urls: string[]` at construction time is the right call — no array checks scattered through the delivery path. Each URL gets its own independent retry cycle, which is correct: a flaky destination shouldn't block retries to a healthy one.

### Missing tests

There are no tests for the fan-out behavior. Given that retry timers are shared across all URLs in a single `Set`, the following scenarios are worth covering:

- An event is delivered to all URLs when multiple are configured.
- A failure for URL A does not prevent delivery to URL B.
- `clearRetryTimers()` on watcher stop cancels pending retries for all URLs.

The existing unit test pattern in `pulse-core.test.ts` uses a mock stream — `pulse-webhooks` will need a similar approach (mock `fetch`).

### Minor note

The retry timer `Set` is shared across all URLs. If URL A fails on attempt 1 and schedules a retry, and the watcher stops before that retry fires, the retry is correctly cancelled. This is the right behaviour — no dangling timers. Just worth documenting in a comment since it's non-obvious that a shared set is safe here.

3m1n3nc3:
> ## Review: Approve with one request
> The fan-out implementation is clean. `ResolvedWebhookConfig` normalising `url` to `urls: string[]` at construction time is the right call — no array checks scattered through the delivery path. Each URL gets its own independent retry cycle, which is correct: a flaky destination shouldn't block retries to a healthy one.
> 
> ### Missing tests
> There are no tests for the fan-out behavior. Given that retry timers are shared across all URLs in a single `Set`, the following scenarios are worth covering:
> 
> * An event is delivered to all URLs when multiple are configured.
> * A failure for URL A does not prevent delivery to URL B.
> * `clearRetryTimers()` on watcher stop cancels pending retries for all URLs.
> 
> The existing unit test pattern in `pulse-core.test.ts` uses a mock stream — `pulse-webhooks` will need a similar approach (mock `fetch`).
> 
> ### Minor note
> The retry timer `Set` is shared across all URLs. If URL A fails on attempt 1 and schedules a retry, and the watcher stops before that retry fires, the retry is correctly cancelled. This is the right behaviour — no dangling timers. Just worth documenting in a comment since it's non-obvious that a shared set is safe here.

I have made all the required changes.

determined-001 (approved):
The three test scenarios correctly cover what was requested:

**"delivers each event to every configured URL"** — verifies fan-out: each event reaches all configured URLs independently, with the correct signature and attempt headers on each request.

**"keeps delivering to other URLs when one URL fails"** — confirms URL isolation: a 500 from one endpoint doesn't suppress delivery to the others, and `webhook.failed` is emitted with the correct `url` and `attempts` fields.

**"cancels pending retries for all URLs when the watcher stops"** — ensures the shared `retryTimers` set does what the new comment says: `watcher.stop()` clears every pending retry regardless of which URL it belongs to.

The `vitest.config.ts` alias resolving `@orbital/pulse-core` to source TypeScript is the right approach — avoids needing a pre-built dist to run the unit tests.

One minor note: `flushAsyncWork` (two `await Promise.resolve()` calls) is a bit fragile if the internal promise chain ever grows a third microtask tick. A `vi.runAllTimersAsync()` or draining the queue in a loop would be more robust, but it works fine for the current implementation. Not a blocker.

determined-001:
Thank you for the contribution! The tests are well-structured and cover all three edge cases we needed. Great work addressing the feedback — really appreciate it! 🎉

PR #239, opened 2026-04-28
fix(122): cap concurrent retries to prevent unbounded memory growth

### LINKED ISSUE #122: 2.16 — Max retry cap to prevent unbounded memory growth

**Context**
Retries push timers into `retryTimers`. With aggressive `retries` config and a perpetually failing receiver, the engine can hold thousands of pending timers.

**Problem**
Unbounded retry queues are a slow-burn memory leak.

**What needs to be done**
1. Cap concurrent in-flight retries per `WebhookDelivery` to a configurable maximum (default 100).
2. When the cap is reached, drop the oldest pending event (or refuse new deliveries — needs design).
3. Emit `webhook.dropped` when an event is dropped due to backpressure.

**Key files**
- `packages/pulse-webhooks/src/index.ts:11` — `retryTimers` set.
- `packages/pulse-webhooks/src/types.ts:1` — config.

**Done when**
- Retry queue is bounded; dropped events emit `webhook.dropped`.

Awointa:
closes #122
- Changed retryTimers from Set to Map<timer, event> to track pending events
- Added maxConcurrentRetries config option (default: 100)
- When cap is reached, oldest pending retry is dropped and webhook.dropped is emitted
- webhook.dropped payload includes reason, url, cap value, and original event

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
       packages: ${{ steps.filter.outputs.packages }}
     steps:
       - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: |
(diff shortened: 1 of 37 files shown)


determined-001:

PR #243, opened 2026-04-28
fix(pulse-core): validate network in EventEngine constructor

### LINKED ISSUE #69: 1.9 — Validate network in EventEngine constructor
**Complexity:** Medium — 150 Points · `type:bug` · `type:dx`

**Context**
`EventEngine` looks up `HORIZON_URLS[config.network]` directly. TypeScript narrows `network` to `"mainnet" | "testnet"`, but a JS consumer or a config loaded from JSON can pass anything.

**Problem**
An unknown network produces `undefined` as the URL, which `Horizon.Server` then accepts and which fails opaquely on the first stream attempt.

**What needs to be done**
1. Add a runtime check: if `HORIZON_URLS[config.network]` is undefined, throw a typed `UnknownNetworkError` listing the valid networks.
2. Export the error class.
3. Add a constructor test.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:45` — constructor.

**Done when**
- Constructing `EventEngine` with a bogus network throws a useful error.
- A test asserts the error type and message.

smartalee:
## Summary
Adds runtime validation to EventEngine constructor to prevent opaque failures when an invalid network is provided.

## Changes
- Added `UnknownNetworkError` class exported from index.ts
- Added runtime check in constructor - throws if network is invalid
- Added tests for network validation

## Testing
- All existing tests pass
- New tests verify error is thrown for invalid network
- Valid networks (mainnet, testnet) work normally

Closes #69


determined-001 (commented):
Thanks @smartalee! The feature code is solid and matches #69 cleanly. Since you're not available, I'm pushing the cleanup directly via maintainer-edit. Here's what I noticed and what I'm fixing:

**What was off:**
1. `packages/pulse-core/package-lock.json` (+2800 lines) was committed — this repo is pnpm-only (root `pnpm-lock.yaml`, no other npm lockfiles anywhere). Looks like `npm install` was run in the package dir instead of `pnpm install` from root.
2. All 3 source files (`EventEngine.ts`, `index.ts`, `pulse-core.test.ts`) end with `\ No newline at end of file` — would create diff noise on future PRs.
3. The branch needs a rebase against `main` (`EventEngine.ts` and `index.ts` have moved).

**What I'm pushing:**
1. Removing the npm lockfile.
2. Restoring trailing newlines on the 3 source files.
3. Rebasing against current `main` and resolving any conflicts.

Your authorship stays on the original commits. Thanks again — the `UnknownNetworkError` class and tests are exactly what #69 asked for.

PR #246, opened 2026-04-28
warn when listeners are added after watcher stop closes #66

### LINKED ISSUE #66: 1.6 — Watcher.on() silently drops handlers after stop

**Context**
After `Watcher.stop()` runs, `_stopped` is set and subsequent `watcher.on(...)` calls return without calling `super.on(...)`.

**Problem**
The handler is silently dropped. A consumer wiring up listeners after stop has no way to know the listener will never fire — a common bug surface during component re-mount races in React consumers.

**What needs to be done**
1. Change `Watcher.on()` so it logs a `console.warn` (or, if a logger is injected per #1.30, uses that) when called on a stopped watcher.
2. Optionally, add a strict mode that throws — gated by a constructor flag.
3. Document the semantics in the README.

**Key files**
- `packages/pulse-core/src/Watcher.ts:16` — `on()` override.

**Done when**
- Adding listeners to a stopped watcher produces a clear warning.
- The README documents the post-stop contract.

Caritajoe18:
fix(pulse-core): warn when listeners are added after watcher stop

Prevent Watcher.on() from failing silently after stop() has been called.
Adding a listener to a stopped watcher now logs a clear warning by
default, and direct Watcher construction can opt into strict mode to
throw instead.

Also document the post-stop listener contract in the pulse-core README
and add tests covering both warning and strict-mode behavior.

closes #66
Screenshot 2026-04-28 at 1 55 21 PM

### DIFF
--- a/packages/pulse-core/README.md
+++ b/packages/pulse-core/README.md
@@ -52,6 +52,8 @@ Open or close the SSE stream. Idempotent — calling `start()` twice logs a warn
 
 Returns a `Watcher` for the given Stellar public key. Watchers are deduplicated — calling `subscribe` twice with the same address returns the same instance.
 
+Once a watcher has been stopped, it will not accept new listeners. Calling `watcher.on(...)` after `watcher.stop()` logs a warning and leaves the listener unregistered. If you construct a watcher directly with `{ strictStoppedListeners: true }`, the same call throws instead.
+
 ### `engine.unsubscribe(address)`
 
 Stops and removes the watcher for the given address.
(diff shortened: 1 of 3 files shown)


determined-001:

determined-001:
Thanks @Caritajoe18! Merged to main as commit 653b80e. Resolved a small conflict with a recent JSDoc pass on Watcher and the new horizonUrl test describe; also tweaked the test to use the injectable logger you added (cleaner than spying on console.warn). Nice touch on the strict-mode option.

Caritajoe18:
@determined-001 
I resolved the conflicts; you can review and merge now

PR #247, opened 2026-04-28
fix: align reconnect attempt numbers across logs and events

### LINKED ISSUE #68: 1.8 — Off-by-one in reconnect log: "attempt N" logged after increment

**Context**
`handleStreamError()` increments `reconnectAttempt` then logs the new value alongside the `engine.reconnecting` notification's `attempt` field.

**Problem**
The `engine.reconnected` notification reports `attempt = pendingReconnectSuccessAttempt`, which is the value *before* increment from the next attempt that succeeded. The numbering is internally consistent but doesn't match the operator's mental model from the warn log line.

**What needs to be done**
1. Audit the attempt-counting flow end-to-end: confirm the warn line, the `engine.reconnecting` payload, and the `engine.reconnected` payload all refer to the same attempt number.
2. Add explicit comments at each emit site explaining what the number represents.
3. Add a test confirming a single retry sequence emits `reconnecting{attempt:1}` then `reconnected{attempt:1}`.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:139` — `handleStreamError()`.
- `packages/pulse-core/src/EventEngine.ts:104` — `pendingReconnectSuccessAttempt` capture.
- `packages/pulse-core/test/pulse-core.test.ts:134` — existing reconnect test to extend.

**Done when**
- All three attempt references for one retry cycle agree on the number.
- The new test verifies the agreement.

Nonso-Eze:
Fixes #68

The reconnect flow now consistently reports the same attempt number across all three touchpoints:
- console.warn log line
- engine.reconnecting notification payload
- engine.reconnected notification payload

Added explicit comments at each emit site explaining what the attempt number represents, and added a test confirming a single retry sequence emits reconnecting{attempt:1} then reconnected{attempt:1}.

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -100,6 +100,8 @@ export class EventEngine {
     this.closeStream();
     this.clearReconnectTimer();
     this.isRunning = true;
+    // Capture the current attempt number for the reconnect success notification.
+    // This value matches the attempt number emitted in engine.reconnecting.
     this.pendingReconnectSuccessAttempt = isReconnect
       ? this.reconnectAttempt
       : null;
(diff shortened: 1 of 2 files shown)


determined-001:

PR #250, opened 2026-04-28
Fix self-payment dedupe in pulse-core

### LINKED ISSUE #67: 1.7 — Self-payment (from === to) fires * twice — dedupe

**Context**
`EventEngine.route()` looks up the destination watcher and emits, then looks up the source watcher and emits. If `from === to` and that address is subscribed, the same payment fires `*` twice with different `type` values.

**Problem**
Self-payments are a small but real percentage of Stellar payments (sweeps, internal rebalances). Consumers building activity feeds get duplicate entries.

**What needs to be done**
1. Detect `event.from === event.to` in `route()` and emit a single `payment.self` typed event instead of two `*` emissions.
2. Add `payment.self` to `PaymentEventType` in `index.ts`.
3. Update the README's event-type table.
4. Add a unit test for the self-payment routing edge case.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:230` — `route()` method.
- `packages/pulse-core/src/index.ts:7` — `PaymentEventType` union.

**Implementation notes**
- Existing consumers that subscribe only to `payment.received` or `payment.sent` will now miss self-payments — call this out in the CHANGELOG.

**Done when**
- Self-payments fire `payment.self` exactly once on the matching watcher.
- A unit test confirms no duplicate `*` emission.

mofejo1:
## Summary
- Route self-payments as a single `payment.self` event.
- Add regression coverage for named and wildcard self-payment emissions.
- Document the new event type and consumer-facing changelog note.

Fixes determined-001/orbital_stellar#67

## Testing
- Not run: Node.js/pnpm are not installed in this environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+
+## Unreleased
+
+- `@orbital/pulse-core`: self-payments where `from === to` now emit a single
+  `payment.self` event instead of separate `payment.received` and
+  `payment.sent` events. Consumers that subscribe only to `payment.received` or
+  `payment.sent` should also subscribe to `payment.self` if they need
+  self-payment activity.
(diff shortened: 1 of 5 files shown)


determined-001:

PR #255, opened 2026-04-28
feat: implement rate limiting handling for Horizon 429 responses in E…

### LINKED ISSUE #88: 1.28 — Client-side rate limiting for Horizon 429 responses
**Complexity:** Medium — 150 Points · `type:perf`

**Context**
Horizon enforces rate limits and returns HTTP 429 when exceeded. The current SSE error handler treats every error identically.

**Problem**
Hammering Horizon with reconnects after a 429 makes the rate limit worse.

**What needs to be done**
1. Detect 429 responses (likely via the SDK error shape — needs investigation).
2. On 429, honor the `Retry-After` header if present, otherwise back off at least 60s before reconnecting.
3. Surface a `engine.rate_limited` notification with the wait duration.

**Key files**
- `packages/pulse-core/src/EventEngine.ts:127` — SSE error handler.
- `packages/pulse-core/src/index.ts:8` — extend `WatcherNotificationType`.

**Done when**
- 429s produce a longer, `Retry-After`-aware delay and a dedicated notification.

Johnsource-hub:
Closes #88 

- Add Horizon `429` rate-limit handling in EventEngine.ts
- Honor `Retry-After` when present, otherwise wait at least `60000ms`
- Emit new watcher notification `engine.rate_limited`
- Update index.ts and docs
- Add tests for `429` with and without `Retry-After`

### DIFF
--- a/packages/pulse-core/README.md
+++ b/packages/pulse-core/README.md
@@ -65,6 +65,7 @@ Stops and removes the watcher for the given address.
 | `*` | `NormalizedEvent` | Any event on this address |
 | `engine.reconnecting` | `WatcherNotification` | The engine is retrying its upstream connection |
 | `engine.reconnected` | `WatcherNotification` | Reconnect succeeded |
+| `engine.rate_limited` | `WatcherNotification` | The engine was rate limited and will retry after the delay |
 
 ### `NormalizedEvent` shape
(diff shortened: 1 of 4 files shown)


determined-001:

PR #433, opened 2026-05-30
Pool pulse-notify EventSource connections

### LINKED ISSUE #432: 3.26 — Connection dedup across hook instances

**Milestone:** v1.0 hardening

**Context**
Today every hook instance opens its own `EventSource`. Two components watching the same address open two connections — wasteful and browser-limit-bound.

**Problem**
The README current-limitations section calls this out.

**What needs to be done**
1. Add an internal connection-pool keyed by `(serverUrl, address, token)`.
2. Multiple hooks share a single `EventSource`; each hook gets its own event filter.
3. Pool closes the connection when the last hook unmounts.

**Key files**
- New: `packages/pulse-notify/src/connectionPool.ts`.

**Done when**
- N components on the same key open 1 EventSource, not N.

newmattock:
## Summary
- add an internal pulse-notify connection pool keyed by serverUrl, address, and token so matching hook instances share one EventSource
- keep per-hook event filtering and state callbacks while closing the pooled connection when the last subscriber unmounts
- document the new connection behavior and add a focused pool regression test

Closes #432

## Verification
- pnpm build
- pnpm test

### DIFF
--- a/apps/web/content/api/pulse-notify.md
+++ b/apps/web/content/api/pulse-notify.md
@@ -147,7 +147,7 @@ The hooks are client-only — they rely on `EventSource`, which doesn't exist in
 
 ## Connection behavior
 
-- One `EventSource` connection per hook instance
+- Hook instances with the same `serverUrl`, `address`, and `token` share one `EventSource`
 - Browser handles reconnection automatically on transient network errors
 - Connections are cleaned up on unmount or when `address` / `serverUrl` change
 - Each render returns the *most recent* event — accumulate history yourself in component state if needed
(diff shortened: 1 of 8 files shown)


determined-001 (approved):
Great work @newmattock — this is exactly what was scoped in #432, nothing more, nothing less. The pool key, shared EventSource, per-subscriber filtering, and last-unmount teardown all land cleanly, and the test asserts the core invariant (N hooks → 1 EventSource) directly.

One small inconsistency I'll clean up in a follow-up: the `test` script uses `tsx test/connectionPool.test.ts` directly rather than the `vitest run` setup the rest of the monorepo uses. I'll add vitest as a devDependency and align the script after merging — no changes needed on your end.

We'll also add you to the contributors list. Welcome to the project! 🎉

PR #436, opened 2026-05-30
fix(pulse-webhooks): support replay clock skew

### LINKED ISSUE #404: 2.67 — Receiver-side replay window: support clock skew
**Complexity:** Trivial — 100 Points · `type:security`

**Milestone:** v1.0 hardening

**Context**
2.35 adds `maxAgeMs`. The check uses `Date.now()`; on a receiver with a skewed clock, valid signatures get rejected.

**Problem**
No allowance for skew.

**What needs to be done**
1. Add `clockSkewMs` (default `30_000`) — accept signatures whose timestamp is up to skew-ms in the future or skew-ms beyond `maxAgeMs` in the past.
2. Document the rationale.

**Key files**
- `packages/pulse-webhooks/src/edge.ts`.

**Done when**
- A timestamp within the skew window is accepted; outside is rejected.

tun1r:
Closes #404.

## What changed
- added `VerifyWebhookOptions` with `maxAgeMs`, `clockSkewMs`, and `nowMs`
- updated `verifyWebhook` to allow bounded future skew and bounded stale skew
- updated `verifyWebhookEdge` with the same replay-window behavior
- documented the replay-window options inline where they are used

## Rationale
Receivers can have slightly skewed clocks. This keeps replay protection strict while allowing a small default tolerance (`clockSkewMs = 30_000`) so valid signatures are not rejected just because the receiver clock is ahead or behind.

## Verification
- `npx vitest run` in `packages/pulse-webhooks` passed
- `npx tsc -p tsconfig.json` in `packages/pulse-webhooks` fails due to pre-existing workspace/type-resolution issues (`@orbital/pulse-core`, Node globals like `Buffer`/`crypto`) not introduced by this change

### DIFF
--- a/packages/pulse-webhooks/src/edge.ts
+++ b/packages/pulse-webhooks/src/edge.ts
@@ -1,5 +1,10 @@
 import type { NormalizedEvent } from "@orbital/pulse-core";
 
+import type { VerifyWebhookOptions } from "./types.js";
+
+const DEFAULT_MAX_AGE_MS = 300_000;
+const DEFAULT_CLOCK_SKEW_MS = 30_000;
+
 /**
  * Verifies webhook signatures using Web Crypto API (compatible with Cloudflare Workers, Deno, and browsers)
  *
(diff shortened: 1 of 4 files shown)


determined-001 (approved):
Great work @tun1r — clean, precise implementation of clock-skew replay protection. The boundary tests are exactly right (±1ms past the limits), pinning `nowMs` in the existing happy-path tests was the right call, and bundling `maxAgeMs` in was necessary since 2.67 depends on it.

One minor quibble I'll clean up as a follow-up: `DEFAULT_MAX_AGE_MS` and `DEFAULT_CLOCK_SKEW_MS` are declared in both `index.ts` and `edge.ts` — I'll consolidate them into `types.ts` alongside `VerifyWebhookOptions` so there's a single source of truth.

We'll also add you to the contributors list. Welcome to the project! 🎉

PR #437, opened 2026-05-30
feat(pulse-webhooks): add custom url validator

### LINKED ISSUE #401: 2.64 — Pluggable URL validator for custom block-lists

**Milestone:** v1.0 hardening

**Context**
The built-in block-list covers private IP ranges. Some consumers want to block more (e.g., known-bad ASNs).

**Problem**
No extension point.

**What needs to be done**
1. Accept `WebhookConfig.urlValidator?: (url) => Promise<string | null>` (returns error message or null).
2. Run after the built-in checks; reject URL if the custom validator returns an error.

**Key files**

**Done when**
- A custom validator rejects URLs the built-in would allow.

tun1r:
Closes #401.

## What changed
- added `WebhookConfig.urlValidator?: (url) => Promise<string | null>`
- run the custom validator during delivery before fetch
- reject delivery when the custom validator returns an error message
- added a test covering a custom validator rejecting a URL the built-in flow otherwise allows

## Scope note
This PR is intentionally minimal and only adds the pluggable validator extension point requested in the issue.

## Verification
- `npx vitest run` in `packages/pulse-webhooks` passed
- `npx tsc -p tsconfig.json` in `packages/pulse-webhooks` fails due to pre-existing workspace/type-resolution issues (`@orbital/pulse-core`, Node globals like `Buffer`/`crypto`) not introduced by this change

### DIFF
--- a/packages/pulse-webhooks/src/index.ts
+++ b/packages/pulse-webhooks/src/index.ts
@@ -5,8 +5,9 @@ import type { WebhookConfig } from "./types.js";
 export { verifyWebhookEdge } from "./edge.js";
 export type { WebhookConfig } from "./types.js";
 
-type ResolvedWebhookConfig = Omit<Required<WebhookConfig>, "url"> & {
+type ResolvedWebhookConfig = Omit<Required<WebhookConfig>, "url" | "urlValidator"> & {
   urls: string[];
+  urlValidator?: WebhookConfig["urlValidator"];
 };
 
 export class WebhookDelivery {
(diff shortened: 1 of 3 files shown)


determined-001 (approved):
Great work @tunirx — clean, minimal implementation exactly as scoped. The `try/catch` around the validator and the `emitFailure` refactor are both the right call.

You're already listed in our contributors from your previous PR — no action needed on your end. Keep the PRs coming! 🙌

PR #460, opened 2026-05-31
feat(abi-registry): implement client with spec-version-negotiation he…

### LINKED ISSUE #365: 1.124 — Registry: spec version negotiation header

**Milestone:** M2 — ABI Registry

**Context**
Future spec format changes need a version-negotiation path so old clients keep working.

**Problem**
The HTTP client does not declare a spec-version header today.

**What needs to be done**
1. Send `Accept: application/vnd.orbital.abi-registry+json; version=1` on every request.
2. Document the version pinning behaviour and how the server should respond.

**Key files**
- `packages/abi-registry/src/AbiRegistryClient.ts`.

**Done when**
- All requests carry the version-pinned `Accept` header.

K1NGD4VID:
closes #365

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -81,6 +81,159 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "brite-side0",
+      "name": "brite-side0",
+      "avatar_url": "https://github.com/brite-side0.png?size=100",
+      "profile": "https://github.com/brite-side0",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
+      "login": "temisan0x",
+      "name": "temisan0x",
+      "avatar_url": "https://github.com/temisan0x.png?size=100",
+      "profile": "https://github.com/temisan0x",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
+      "login": "MAN7A-afk",
+      "name": "MAN7A-afk",
+      "avatar_url": "https://github.com/MAN7A-afk.png?size=100",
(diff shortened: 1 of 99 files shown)


determined-001 (changes requested):
Thanks for the spec version negotiation header (#365)! Please remove `packages/pulse-notify/test/connectionPool.test.ts` — it's unrelated to the ABI registry work. Remove it and we'll merge.

K1NGD4VID:
fixed please review, and thanks for the issue assignment

determined-001 (approved):
## Review — matches issue #365 ✅

Confirmed against the issue's acceptance criteria:

- **"Send `Accept: application/vnd.orbital.abi-registry+json; version=1` on every request"** — `getSpec()` and `getSpecs()` both route through the single private `fetchBatch()` → `POST /specs` call, and the header is added there, so *every* network request the client makes carries the version-pinned `Accept` header. ✓
- **"Document the version pinning behaviour and how the server should respond"** — the JSDoc covers client pinning, the 200 (supported) and 406 Not Acceptable (unsupported) server responses, and forward-compatibility. ✓
- `REGISTRY_SPEC_VERSION = 1` is a clean single source of truth for the version. ✓

Scope is tight and on-target. The only change beyond the header is removing a redundant duplicate `RegistryPublisher` export from the `export type` block in `index.ts` (it's still exported on line 3) — harmless de-dup.

Thank you @K1NGD4VID — this is exactly the kind of small, well-documented, precisely-scoped PR that's a pleasure to review. The 406 documentation in particular is a nice touch. 🙏

PR #495, opened 2026-05-31
Add optional subscription names for observability in pulse-core

### LINKED ISSUE #323: 1.82 — engine.subscribeContract accepts a name?: string for observability
**Complexity:** Trivial — 100 Points · `type:dx`

**Milestone:** v1.0 hardening

**Context**
Multiple contract subscriptions look identical in logs without a friendly identifier.

**Problem**
Operators cannot tell which subscription a log line refers to.

**What needs to be done**
1. Add optional `name?: string` to `ContractSubscriptionConfig`.
2. Include it in every log line and lifecycle notification.

**Key files**

**Done when**
- A named subscription appears with its name in log output and notifications.

temisan0x:
Closes #323

## What changed
- Added name?: string to SubscribeOptions in packages/pulse-core/src/index.ts
- Exported ContractSubscriptionConfig as an alias for SubscribeOptions
- Extended WatcherNotification with optional name?: string
- Tracked subscription names per address in EventEngine with a subscriptionNames map
- Threaded the name into lifecycle notifications and subscription warnings via describeSubscription()
- Added test coverage for named subscriptions and labeled notifications
## Note on scope
The issue refers to engine.subscribeContract, but that API does not exist as a separate method in this codebase. The observability label was therefore applied to the existing subscribe() flow, which is the actual subscription entry point used by pulse-core.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -63,6 +63,168 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "newmattock",
+      "name": "newmattock",
+      "avatar_url": "https://github.com/newmattock.png?size=100",
+      "profile": "https://github.com/newmattock",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
+      "login": "tunirx",
+      "name": "Tunir",
+      "avatar_url": "https://github.com/tunirx.png?size=100",
+      "profile": "https://github.com/tunirx",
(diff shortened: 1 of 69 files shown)


determined-001 (approved):
Great work @temisan0x — minimal and precise name option on subscribeContract for issue #323. You're already listed in our contributors. 🎉

PR #505, opened 2026-05-31
Fix shared connected state for pulse-notify event connection pool

### LINKED ISSUE #419: 3.37 — EventState.connected reflects pool-level state
**Complexity:** Trivial — 100 Points · `type:refactor`

**Milestone:** v1.0 hardening

**Context**
After 3.26, `connected` should reflect the pool's connection state, not the hook's.

**Problem**
Without this, two hooks sharing a connection see drift between their `connected` values.

**What needs to be done**
1. Wire `connected` to read from the pool entry.
2. Update all hooks on pool state change.

**Key files**
- `packages/pulse-notify/src/connectionPool.ts`.

**Done when**
- Two hooks sharing a connection see the same `connected` value.

temisan0x:
### Summary

Ensure useStellarEvent hooks share the same live connection state when they reuse a pooled SSE connection.
### What changed
- connectionPool.ts
  - acquireEventConnection() now returns connected as a getter so it reflects the current shared pool entry state instead of a stale snapshot.
- index.ts
  - Cleaned up merge conflict artifacts and restored the stable useStellarEvent implementation.
- connectionPool.test.ts
  - Added regression coverage verifying two subscribers on the same pooled connection both see connected = true after onopen.
### Why
Without this fix, multiple hooks sharing the same underlying connection could drift on whether the connection was considered open, causing inconsistent UI/status behavior across hooks using the same event stream.

Closes #419

### DIFF
--- a/packages/pulse-notify/src/connectionPool.ts
+++ b/packages/pulse-notify/src/connectionPool.ts
@@ -80,7 +80,9 @@ export function acquireEventConnection(
   entry.subscribers.add(subscriber);
 
   return {
-    connected: entry.connected,
+    get connected() {
+      return entry.connected;
+    },
     unsubscribe: () => {
       entry.subscribers.delete(subscriber);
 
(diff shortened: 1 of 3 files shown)


determined-001 (approved):
Great work @temisan0x — clean pool-level connected state wiring for issue #419. You're already listed in our contributors. 🎉

PR #528, opened 2026-05-31
Add retry queue visibility timeout and ack/nack handling

### LINKED ISSUE #380: 2.43 — Replay queue: visibility timeout for in-flight records

**Milestone:** M4 — Replay primitives

**Context**
Dequeue marks the record as in-flight; if the worker crashes before ack, the record should re-emerge.

**Problem**
The current interface has no concept of in-flight.

**What needs to be done**
1. Add `ack(recordId)` and `nack(recordId, requeueDelayMs)` to the interface.
2. Default visibility timeout 30s, configurable per adapter.
3. Adapters that have native visibility (SQS) use it; others (Postgres) implement via a column.

**Key files**
- `packages/pulse-webhooks/src/RetryQueue.ts`.
- All four adapter files.

**Done when**
- A dequeued-but-not-acked record re-emerges after the visibility timeout.

Anthony-19:
Closes #380

### DIFF
--- a/packages/pulse-webhooks/src/RedisRetryQueue.ts
+++ b/packages/pulse-webhooks/src/RedisRetryQueue.ts
@@ -28,18 +28,22 @@ export type RedisRetryQueueOptions = {
   queueName?: string;
   now?: () => number;
   scanBatchSize?: number;
+  visibilityTimeoutMs?: number;
 };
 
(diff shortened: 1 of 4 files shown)


determined-001 (approved):
Solid visibility timeout + ack/nack implementation (#380). Clean RetryQueue extension. Thanks for the contribution!

PR #532, opened 2026-06-01
feat(pulse-core): add cacheCursorStore LRU/TTL wrapper (#344)

### LINKED ISSUE #344: 1.103 — CursorStore LRU read cache wrapper
**Complexity:** Trivial — 100 Points · `type:perf`

**Context**
On engine restart, the same cursor is read multiple times before the first event arrives.

**Problem**
Repeated reads against Postgres / S3 are unnecessary.

**What needs to be done**
1. Add `cacheCursorStore(inner, { ttlMs }): CursorStore` with an LRU cache over reads.
2. Invalidate cache entries on `set`.

**Key files**
- New: `packages/pulse-core/src/cacheCursorStore.ts`.

**Done when**
- A wrapped store reads the inner store at most once per key per `ttlMs` window.

JoesWalker:
Description:
  
  ## Summary
  
  Closes #344.
  
  Adds `cacheCursorStore(inner, { ttlMs })` — a thin TTL cache wrapper around any `CursorStore` implementation.
  
  - `get`: serves from an in-memory `Map` if the entry is still within TTL; otherwise delegates to `inner` and caches the result
  - `set`: invalidates the cache entry for the key, then delegates to `inner`
  - No new dependencies — minimal inline implementation using `Map` + `Date.now()`
  - Exported from `@orbital/pulse-core` public index
  
  ## Files changed
  
  - `packages/pulse-core/src/cacheCursorStore.ts` — implementation
  - `packages/pulse-core/test/cacheCursorStore.test.ts` — 3 unit tests covering: cache hit within TTL, re-fetch after TTL expiry, re-fetch after `set`
  invalidation
  - `packages/pulse-core/src/index.ts` — export added
  
  ## Testing
  
  pnpm --filter @orbital/pulse-core exec vitest run cacheCursorStore
  
  3 tests passed

### DIFF
--- a/packages/pulse-core/src/cacheCursorStore.ts
+++ b/packages/pulse-core/src/cacheCursorStore.ts
@@ -1,0 +1,33 @@
+import type { CursorStore } from "./CursorStore.js";
+
+interface Entry {
+  value: string | null;
+  expiresAt: number;
+}
+
+/**
+ * Wraps a CursorStore with an in-memory TTL cache.
+ * - get: returns cached value if still fresh, otherwise delegates to inner and caches result.
+ * - set: invalidates the cache entry, then delegates to inner.
+ */
+export function cacheCursorStore(
+  inner: CursorStore,
+  { ttlMs }: { ttlMs: number },
+): CursorStore {
+  const cache = new Map<string, Entry>();
+
+  return {
+    async get(streamKey: string): Promise<string | null> {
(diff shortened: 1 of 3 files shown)


determined-001 (approved):
Clean `cacheCursorStore` TTL wrapper (#344). Minimal implementation with the right delegation pattern. Thanks for the contribution!

PR #545, opened 2026-06-01
 3.45 — Server-side fallback: render last-known event during SSR

### LINKED ISSUE #427: 3.45 — Server-side fallback: render last-known event during SSR

**Milestone:** v1.0 hardening

**Context**
SSR delivers no event. A common pattern: SSR fetches the last-known event from a server endpoint and passes it as initial state.

**Problem**
No standard way to pass initial state.

**What needs to be done**
1. Accept `initialEvent?: T | null` on every hook.
2. Use it as the initial state; replace on first live event.

**Key files**

**Done when**
- A hook with `initialEvent` renders that event during SSR and switches to live after hydration.

brite-side0:
Closes #427  3.45 — Server-side fallback: render last-known event during SSR
…arithmetic

The remote implementation used BigInt(Math.round(parseFloat(amount) * 10_000_000)) which reintroduces floating-point precision loss for amounts like 0.0000001.

Replace with amountToStroop() that splits on the decimal point and pads the fractional part to exactly 7 digits before combining with BigInt safe for all amounts representable on the Stellar network.

## Linked issue

Closes #

## Test plan

## Breaking changes

None / Yes (describe below)

### DIFF
--- a/packages/pulse-notify/src/index.ts
+++ b/packages/pulse-notify/src/index.ts
@@ -127,6 +127,24 @@ export function useStellarEvent<T extends NormalizedEvent = NormalizedEvent>(
 
 type PaymentEvent = Extract<NormalizedEvent, { type: "payment.received" }>;
 
+/**
+ * Converts a Stellar decimal amount string (e.g. "12.3456789") to stroops
+ * (1 XLM = 10,000,000 stroops) as a bigint.
+ *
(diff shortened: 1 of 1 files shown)


determined-001 (approved):
Clean SSR fallback implementation (#427). `initialEvent` option is the right API shape. Thanks!

PR #557, opened 2026-06-02
Add: a decoded: Uint8Array | null sibling to DataEvent.value

### LINKED ISSUE #332: 1.91 — Union refinement: data.value decoded shape
**Complexity:** Trivial — 100 Points · `type:refactor`

**Context**
`DataEvent.value` is `string | null` — base64-encoded by Horizon. Consumers always decode to compare.

**Problem**
Every consumer writes the same `Buffer.from(value, 'base64')` boilerplate.

**What needs to be done**
1. Add a `decoded: Uint8Array | null` sibling to `DataEvent.value`.
2. Document `value` as the raw base64 string and `decoded` as the bytes.

**Key files**
- `packages/pulse-core/src/EventEngine.ts` (normalizer).

**Done when**
- Every `DataEvent` carries both `value` and `decoded`.

Dydex:
closes #332

DataEvent.value already exposed the raw Horizon base64 string for manage-data operations, but consumers had to decode it themselves every time. This PR adds a decoded field alongside value populated at normalisation time by Buffer.from(value, "base64"), falling back to null if the value is absent or the base64 is malformed. The DataEvent type in index.ts is updated with JSDoc comments distinguishing the two fields. No breaking changes  value is untouched.

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -613,11 +613,21 @@ export class EventEngine {
     const value = r.data_value == null ? null : String(r.data_value);
     const type: DataEventType = value !== null ? "data.set" : "data.cleared";
 
+    let decoded: Uint8Array | null = null;
+    if (value !== null) {
+      try {
+        decoded = Buffer.from(value, "base64");
+      } catch (err) {
+        decoded = null;
+      }
+    }
+
     return {
       type,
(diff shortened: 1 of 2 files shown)


determined-001 (approved):
Clean `decoded: Uint8Array | null` addition to DataEvent (#332). Exactly what the issue asked for. Thanks!

PR #558, opened 2026-06-02
Add: EventEngine.subscribeContract returns Promise for first-event readiness

### LINKED ISSUE #319: 1.78 — EventEngine.subscribeContract returns Promise for first-event readiness
**Complexity:** Medium — 150 Points · `type:dx`

**Milestone:** v1.0 hardening

**Context**
After `subscribeContract`, the first event arrives whenever the next poll completes. There is no signal for "subscription is now active."

**Problem**
Tests racing the first event are flaky.

**What needs to be done**
1. Add `engine.awaitContractSubscriptionActive(config, { timeoutMs }): Promise<void>`.
2. Resolves when the subscriber has issued the first `getEvents` call that includes the new filter.
3. Rejects on timeout.

**Key files**
- `packages/pulse-core/test/EventEngine.subscribeContract.test.ts`.

**Done when**
- The promise resolves after the first poll covering the filter completes.

Dydex:
closes #319 

Adds awaitContractSubscriptionActive() to EventEngine — a Promise-based API that resolves when the engine has confirmed it is actively polling a given contract+topics combination. Internally, notifyContractPolled() signals each poll cycle; the awaiter checks whether the polled topics fully cover the requested set and resolves if so. A configurable timeoutMs rejects the promise if readiness isn't confirmed in time. This lets callers await before sending on-chain transactions, removing the race condition where events fire before the subscription is live. 4 vitest cases cover: topic superset match, wildcard (no-topic) poll, partial-topic timeout, and no-topics-requested fast path.

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -85,6 +85,16 @@ export class EventEngine {
   private readonly reconnectConfig: Required<ReconnectConfig>;
   private isRunning = false;
   private filters: Map<string, (event: NormalizedEvent) => boolean> = new Map();
+  // Waiters for contract subscription activation: map contractId -> array of waiters
+  private contractPollWaiters: Map<
+    string,
+    Array<{
+      topics?: string[] | undefined;
+      resolve: () => void;
+      reject: (err: unknown) => void;
+      timeout?: ReturnType<typeof setTimeout> | null;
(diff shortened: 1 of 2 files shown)


determined-001 (approved):
Clean `awaitContractSubscriptionActive()` implementation (#319). Two focused files. Thanks!

PR #559, opened 2026-06-02
Add: S3CursorStore reference adapter

### LINKED ISSUE #340: 1.99 — S3CursorStore reference adapter

**Context**
S3 is the cheapest durable store for low-write-rate cursors.

**Problem**
No S3 adapter ships today.

**What needs to be done**
1. Add `S3CursorStore` accepting an `S3Like` interface (`getObject`, `putObject`).
2. One object per stream key under `cursors/<streamKey>.json`.
3. Tolerate `NoSuchKey` as `null`; treat any other error as a failure.

**Key files**
- New: `packages/pulse-core/src/S3CursorStore.ts`.
- New: `packages/pulse-core/test/S3CursorStore.test.ts`.

**Done when**
- The adapter round-trips against a mock S3 client; the interface compiles without a real AWS SDK dep.

Dydex:
closes #340

This Pr Introduces S3CursorStore, a CursorStore implementation that persists stream cursors to an S3-compatible bucket. It depends on an S3Like interface (duck-typed, so it works with AWS SDK v2, v3, MinIO, etc.) rather than a concrete SDK. Cursors are stored as { cursor: string } JSON objects under a configurable key prefix (cursors/<streamKey>.json). getCursor() gracefully handles NoSuchKey by returning null (first-run safe), while re-throwing all other S3 errors. putCursor() serialises and writes atomically via putObject. Tests use an in-memory mock S3 and cover round-trips, missing-key → null, and non-NoSuchKey error propagation.

### DIFF
--- a/packages/pulse-core/src/S3CursorStore.ts
+++ b/packages/pulse-core/src/S3CursorStore.ts
@@ -1,0 +1,64 @@
+export type S3Like = {
+  getObject(params: {
+    Bucket: string;
+    Key: string;
+  }): Promise<{ Body: string | Uint8Array }>;
+  putObject(params: {
+    Bucket: string;
+    Key: string;
+    Body: string | Uint8Array;
+  }): Promise<void>;
+};
+
+export class S3CursorStore {
+  private readonly s3: S3Like;
+  private readonly bucket: string;
+  private readonly prefix: string;
+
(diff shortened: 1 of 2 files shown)


determined-001 (approved):
Solid `S3CursorStore` adapter (#340) — clean `S3Like` duck-typed interface is the right approach. Thanks!

PR #718, opened 2026-06-26
Soroban benchmark 2

samuelfrancis163-eng:
Closes #617 

## Test plan

- [ ] Existing tests pass (`pnpm test`)
- [ ] New tests added for new behaviour
- [ ] Typechecks pass (`pnpm -r typecheck`)
- [ ] Manually tested against testnet / mainnet (describe what you tested)

## Breaking changes

None / Yes (describe below)

PR #719, opened 2026-06-26
feat: normalize contract event shapes and align with public types

samuelfrancis163-eng:
This PR updates the event normalizer function `normalizeContractEvent` in `packages/pulse-core/src/EventEngine.ts` to return the official, public `ContractInvokedEvent` and `ContractEmittedEvent` types as defined in `packages/pulse-core/src/index.ts`. 

Previously, the normalizer returned internal wrapper types (`RpcContractInvokedEvent` / `RpcContractEmittedEvent`) using type identifiers like `contract_invoked` and `contract_emitted`.

### Key Changes
1. **Domain Alignment**: Updated the normalizer logic to output `type: "contract.invoked"` and `type: "contract.emitted"`.
2. **Field Mapping**: Mapped all expected domain event properties:
   - `eventId` (derived from RPC event `id`).
   - `timestamp` (derived from RPC event `ledgerClosedAt`).
   - `ledger` and `txHash`.
3. **Backward Compatibility**: Preserved legacy properties (`id`, `pagingToken`) on the resolved JS objects to prevent breaking any consumer usages or test assertions expecting them.
4. **Tests Updates**:
   - Updated `EventEngine.normalizeContract.test.ts` and `SorobanRpcClient.xdrFormat.test.ts` assertions to check for the correct `contract.emitted` type.
   - Fixed type assertions in `types.contract.test.ts` to ensure the `raw` property is populated on mock objects, avoiding runtime typecheck errors in newer Vitest releases.


determined-001 (changes requested):
Same issue as #718 from the same branch: `packages/pulse-core/src/pr_placeholder.txt` must be removed, and the `package.json` change needs to not overwrite `format:all`. Please also add a `Closes #NNN` line to the PR body so we can link the linked issue. Once cleaned up and rebased on current main, resubmit.

determined-001 (changes requested):
The PR includes `packages/pulse-core/src/pr_placeholder.txt` (a placeholder file that should not be committed) and a root `package.json` change that is out of scope. Please remove these and resubmit only the implementation for the linked issue.

PR #720, opened 2026-06-26
Cursor store health check integration

### LINKED ISSUE #644: 1.77 — CursorStore health check integration

**Context**
A wedged cursor store is a real outage cause; liveness probes should surface it.

**Problem**
No standard probe for the cursor store.

**What needs to be done**
1. Add optional `ping(): Promise<void>` to `CursorStore`.
2. `engine.healthCheck()` includes a cursor-store ping result when configured.

**Key files**
- `packages/pulse-core/src/CursorStore.ts`

**Done when**
- A failing `ping` flips `engine.healthCheck().ok` to `false`.

Mikey-222:
## Linked issue

Closes #644

## Test plan

## Breaking changes

None / Yes (describe below)

### DIFF
--- a/apps/web/app/starters/page.tsx
+++ b/apps/web/app/starters/page.tsx
@@ -1,0 +1,5 @@
+import StartersPage from '@/components/StartersPage'
+
+export default function Page() {
+  return <StartersPage />
+}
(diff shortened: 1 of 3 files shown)


determined-001:
Thanks @Mikey-222! The `ping?: () => Promise<void>` type fix is correct — the return value is discarded by the health check, so `void` is the right contract. The starters page files that arrived via an older base commit are already on main from another PR, so the squash only includes the CursorStore change. Merging now — closes #644.

PR #743, opened 2026-06-27
healthCheck: check Soroban subscriber alongside Horizon

### LINKED ISSUE #638: 1.71 — engine.healthCheck() for liveness probes

**Milestone:** v1.0 hardening

**Context**
Kubernetes / Vercel probe URLs need a fast, side-effect-free liveness signal.

**Problem**
`engine.status()` is informational, not actionable.

**What needs to be done**
1. Add `engine.healthCheck(): { ok: boolean; reasons: string[] }`.
2. `ok = true` only if every active source is running and `lastEventAt` is within a configurable threshold (default 5 min).
3. `reasons` lists failure causes.

**Key files**
- New: `packages/pulse-core/test/EventEngine.healthCheck.test.ts`

**Done when**
- A wedged source (no events in > threshold) flips `ok` to `false` with a clear reason.

nice-bills:
Closes #638 
healthCheck() now checks both Horizon and Soroban subscriber sources. Each source's running state and lastEventAt staleness are validated against a configurable threshold (default 5 min), with scoped failure reasons per source.

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -578,19 +578,37 @@ export class EventEngine {
 
   async healthCheck(thresholdMs = 5 * 60 * 1000): Promise<HealthCheckResult> {
     const reasons: string[] = [];
+
     if (!this.isRunning) {
-      reasons.push("engine is not running");
+      reasons.push("horizon source is not running");
     }
     if (this.lastEventAt === null) {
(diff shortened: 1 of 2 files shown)


determined-001:
Thank you, @nice-bills! Extending healthCheck() to include Soroban subscriber liveness alongside Horizon, with the updated reason message prefixes and comprehensive tests, is exactly what issue #638 asked for. Merging now.

PR #753, opened 2026-06-27
perf(pulse-core): coalesce duplicate filters into the minimum getEvents calls (#634)

### LINKED ISSUE #634: 1.67 — Soroban subscriber: coalesce active filters across subscriptions

**Context**
Each `subscribeContract` causes a separate poll. N subscriptions = N polls per interval.

**Problem**
Wasteful at scale; multiple subscriptions share most ledgers.

**What needs to be done**
1. Coalesce active filters into minimum RPC calls (up to 5-filter cap per call).
2. Route returned events back to originating subscriptions via per-subscription matching.
3. N subscriptions ≤ 5 → one RPC call; N > 5 → multiple calls within the 5-filter budget.

**Key files**
- New: `packages/pulse-core/test/SorobanSubscriber.coalesce.test.ts`

**Done when**
- N ≤ 5 subscriptions produce one RPC call per poll. N > 5 uses the minimum number of calls.

Ipramking:
## Summary

Implements **#634 (1.67 — Soroban subscriber: coalesce active filters across subscriptions)**.

`SorobanSubscriber._doPoll` already flattened all subscription filters and chunked them under the 5-filter-per-call cap. The remaining gap for *minimum* RPC calls: **duplicate filters across subscriptions each consumed a filter slot**, so N subscriptions watching the same contract cost `ceil(N / 5)` calls instead of one.

## What changed
- Added `coalesceFilters()` — collapses identical filters (order-preserving) before chunking. Two filters are identical when they share the same `type`, the same **set** of contract IDs (order-independent, via a sorted key), and the same positional `topicFilters`.
- `_doPoll` now chunks the **deduplicated** filters under the 5-filter cap.
- Routing is unchanged: returned events are still dispatched to **every** subscription whose filters match (per-subscription matching), so coalescing never drops a delivery.

## Done when ✓
- **N ≤ 5 → one RPC call per poll**, and **N > 5 → minimum number of calls** (duplicates no longer inflate the count).
  - 8 subscriptions across 2 contracts → **1** call (was 2).
  - 12 subscriptions across 6 distinct contracts → **2** calls (`ceil(6/5)`, was 3).

## Tests (`SorobanSubscriber.coalesce.test.ts`)
Added 4 cases: duplicate-filter coalescing to one slot, contract-id-order treated as identical, minimum-calls when unique filters exceed the cap, and a coalesced event delivered to every matching subscription. All 9 coalesce tests pass; full `pulse-core` suite green (**444 passed, 6 skipped**); `tsc -p` build, `test:types`, ESLint, and Prettier all clean.

Closes #634

### DIFF
--- a/packages/pulse-core/src/SorobanSubscriber.ts
+++ b/packages/pulse-core/src/SorobanSubscriber.ts
@@ -315,11 +315,17 @@ export class SorobanSubscriber {
         flatFilters.push(...sub.filters);
       }
 
-      if (flatFilters.length === 0) {
+      // Coalesce identical filters (order-preserving) so duplicate subscriptions
+      // share a single filter slot. This minimises the number of getEvents calls
+      // under the 5-filter cap — e.g. 6 subscriptions on the same contract collapse
+      // to one filter, hence one call instead of two.
+      const uniqueFilters = this.coalesceFilters(flatFilters);
+
+      if (uniqueFilters.length === 0) {
         rpcCalls = [[]];
       } else {
-        for (let i = 0; i < flatFilters.length; i += 5) {
-          rpcCalls.push(flatFilters.slice(i, i + 5));
+        for (let i = 0; i < uniqueFilters.length; i += 5) {
+          rpcCalls.push(uniqueFilters.slice(i, i + 5));
         }
       }
     }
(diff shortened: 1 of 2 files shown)


determined-001:
Thank you, @Ipramking! Filter coalescing to minimize getEvents calls under the 5-filter RPC cap is exactly what issue #634 asked for. The stable identity key and FIFO ordering make this a clean implementation. Merging now.

PR #762, opened 2026-06-27
Fix/#655

### LINKED ISSUE #655: 1.88 — Typed amount as StellarAmount branded string
**Complexity:** Medium — 150 Points · `type:refactor`

**Context**
Every monetary `amount: string` field is documented as "decimal string; do not use
JS numbers." The type system does not enforce this.

**Problem**
Consumers occasionally do `Number(event.amount)`, silently losing precision.

**What needs to be done**
1. Introduce branded type `StellarAmount = string & { __brand: 'StellarAmount' }`.
2. Narrow every event's `amount` to `StellarAmount`.
3. Export `toBigInt(amount: StellarAmount): bigint` (multiplied by 10^7 for stroop precision).

**Key files**
- New: `packages/pulse-core/src/amount.ts`
- `packages/pulse-core/src/index.ts`
- New: `packages/pulse-core/test/amount.test.ts`

**Done when**
- `Number(event.amount)` requires an explicit cast. `toBigInt(event.amount)` returns the stroop-precision integer.

AYo101o:
Closes #655

### DIFF
--- a/packages/pulse-core/src/index.ts
+++ b/packages/pulse-core/src/index.ts
@@ -16,6 +16,7 @@ export type {
 
 export { validateContractFilters } from "./contractFilters.js";
 export { Watcher } from "./Watcher.js";
+export { toStellarAmount, toBigInt } from "./amount.js";
 export type { StellarAmount } from "./amount.js";
 export type { AccountAddress, MuxedAddress, ContractAddress } from "./address.js";
 export { EngineAlreadyStartedError, HorizonStreamError } from "./errors.js";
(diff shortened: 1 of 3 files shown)


determined-001:
Thanks @AYo101o! Exporting `toStellarAmount` and `toBigInt` from the package index is the right fix for issue #655 — the branded type was already defined in `amount.ts` but wasn't reachable from the package entry point. Removed the stray file-header comment from `RedisRetryQueue.ts`. Merging now.

PR #765, opened 2026-06-27
closes #651 1.84 — TypeScript type generation from spec (codegen CLI) 

dev-sylv:
Closes #651 1.84 — TypeScript type generation from spec (codegen CLI)

- [ ] Existing tests pass (`pnpm test`)
- [ ] New tests added for new behaviour
- [ ] Typechecks pass (`pnpm -r typecheck`)
- [ ] Manually tested against testnet / mainnet (describe what you tested)

PR #767, opened 2026-06-27
Handle cursor-expired errors on Soroban subscriber resume

### LINKED ISSUE #626: 1.59 — Handle cursor-expired errors on Soroban subscriber resume

**Context**
Stellar RPC retains 24 h of events by default (7 d max). A consumer stopped longer
than retention has a cursor pointing past the visible window.

**Problem**
The subscriber would fail in a loop on resume.

**What needs to be done**
1. Catch the cursor-out-of-range error from `getEvents`.
2. Emit `engine.cursor_expired` with `{ source, lostCursor }`.
3. Fall back to `startLedger = latestLedger`; warn log about data loss.

**Key files**
- `packages/pulse-core/src/index.ts`
- New: `packages/pulse-core/test/SorobanSubscriber.cursorExpired.test.ts`

**Done when**
- A subscriber with an expired cursor recovers, emits `engine.cursor_expired`, and continues from `latestLedger`.

Emelie-Dev:
closes #626

### DIFF
--- a/packages/pulse-core/src/SorobanSubscriber.ts
+++ b/packages/pulse-core/src/SorobanSubscriber.ts
@@ -162,6 +162,7 @@ export class SorobanSubscriber extends EventEmitter {
   private retryTimer: ReturnType<typeof setTimeout> | null = null;
 
   constructor(options: SorobanSubscriberOptions) {
+    super();
     const pageLimit = options.pageLimit ?? options.pageSize ?? DEFAULT_PAGE_LIMIT;
     if (!Number.isFinite(pageLimit) || pageLimit < MIN_PAGE_LIMIT || pageLimit > MAX_PAGE_LIMIT) {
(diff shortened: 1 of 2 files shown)


Emelie-Dev:
@determined-001 

Please i do not think the CI failures are caused by my changes. I've verified locally that 
upstream `main` has `zod@^4.4.3` in `packages/abi-registry/package.json` 
(added in #765) but `pnpm-lock.yaml` was never updated to match. Running 
`pnpm install --frozen-lockfile` against upstream `main` directly reproduces 
the error. Any PR branched from current `main` will fail CI until the upstream 
lockfile is regenerated and committed.

determined-001:
Excellent implementation of cursor-expired recovery for issue #626 — detects the `invalid_request`/`startCursor` RPC error, emits `engine.cursor_expired`, falls back to the latest ledger, and advances the cursor to avoid an infinite retry loop. All CI checks green. Merging now, thank you @Emelie-Dev!

PR #776, opened 2026-06-27
fixed issues

### LINKED ISSUE #689: 3.26 — Connection dedup across hook instances

**Milestone:** v1.0 hardening

**Context**
Today every hook instance opens its own `EventSource`. Two components watching the same address open two connections.

**Problem**
Wasteful; browser connection-per-host limits apply.

**What needs to be done**
1. Add internal connection-pool keyed by `(serverUrl, address, token)`.
2. Multiple hooks share a single `EventSource`; each gets its own event filter.
3. Pool closes the connection when the last hook unmounts.

**Key files**
- New: `packages/pulse-notify/src/connectionPool.ts`
- `packages/pulse-notify/src/index.ts`

**Done when**
- N components on the same key open 1 EventSource, not N.

Emelie-Dev:
closes #689

### DIFF
--- a/packages/pulse-notify/src/StellarConnectionStatus.tsx
+++ b/packages/pulse-notify/src/StellarConnectionStatus.tsx
@@ -1,6 +1,7 @@
 import { createElement, useEffect, useMemo, useState } from "react";
 import { StellarEventBoundary } from "./StellarEventBoundary.js";
 import type { ComponentPropsWithoutRef, CSSProperties, ReactElement } from "react";
+import { acquireEventConnection } from "./connectionPool.js";
 
 export type StellarConnectionStatusState = "connecting" | "connected" | "error";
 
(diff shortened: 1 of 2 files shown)


determined-001 (changes requested):
Thanks @Emelie-Dev — the `url-validator.ts` cleanup (removing unused catch binding, narrowing the `data` type, `??` over `||`) is correct code, but it doesn't implement issue #689 ("Connection dedup across hook instances"), which is a `pulse-notify` feature about sharing a single SSE connection across multiple hook instances subscribing to the same address. Please either: (a) link the correct issue that this cleanup addresses, or (b) implement the actual dedup logic and push it here.

determined-001:
This correctly routes `StellarConnectionStatus` through the shared connection pool (via `acquireEventConnection`) instead of opening its own raw `EventSource` — that's the dedup for issue #689. All CI checks pass. Merging now, thank you @Emelie-Dev!

PR #783, opened 2026-06-27
feat(pulse-core): LRU dedup window for Soroban events, default 1024 (#635)

### LINKED ISSUE #635: 1.68 — Soroban subscriber: duplicate event dedup via LRU

**Context**
Under retry, the same event ID can be returned twice from the RPC.

**Problem**
Downstream consumers see duplicate `contract.emitted` events.

**What needs to be done**
1. Maintain a small LRU of recent event IDs (default 1024).
2. Skip emission if the event ID was seen recently.
3. Document the cap and dedup-window guarantee.

**Key files**
- New: `packages/pulse-core/test/SorobanSubscriber.dedup.test.ts`

**Done when**
- Replaying the same page twice produces events exactly once downstream.

chizzy0011:
## Summary

Implements **#635 (1.68 — duplicate event dedup via LRU)**.

The SorobanSubscriber already de-duplicated events across polls, but with a **FIFO** `Set` (default 10,000). Under retry the same event ID can reappear far apart; a FIFO window can evict an ID that is still actively being re-seen. This converts it to a **true LRU**.

## What changed (`SorobanSubscriber.ts`)
- **LRU window** — re-seeing an ID now **refreshes its recency** (`touchSeen`, on a dedup hit), and `recordSeen` evicts the **least-recently-used** ID when the cap is exceeded. A frequently-repeated ID therefore stays de-duplicated instead of aging out by insertion order.
- **Default 1024** — `DEFAULT_DEDUP_CACHE_SIZE` is now 1024 (was 10,000), per the issue. Still overridable via `dedupCacheSize`.
- **Docs** — the `dedupCacheSize` option, the `seen` field, and `recordSeen`/`touchSeen` now document the cap and the dedup-window guarantee: *an event ID is delivered downstream at most once while it remains in the window.*

## Done when ✓
- Replaying the same page twice produces each event exactly once downstream (existing + new tests).

## Tests (`SorobanSubscriber.dedup.test.ts`)
Kept the existing coverage and added two LRU-specific cases:
- **Refresh-on-repeat (LRU, not FIFO)** — re-seeing an ID keeps it and evicts a different, less-recently-used ID; the assertion fails under the old FIFO behaviour.
- **Default window is 1024** — 1025 distinct IDs evict the first (re-emittable) while the newest stays suppressed.

## Verification
- `tsc -p` build + `test:types` — clean.
- `vitest` dedup file — 11/11; full `pulse-core` suite green: **480 passed / 7 skipped**.
- Prettier + ESLint clean.

Closes #635

### DIFF
--- a/packages/pulse-core/src/SorobanSubscriber.ts
+++ b/packages/pulse-core/src/SorobanSubscriber.ts
@@ -84,7 +84,12 @@ export interface SorobanSubscriberOptions {
   pageLimit?: number;
   /** @deprecated Alias for {@link SorobanSubscriberOptions.pageLimit}. */
   pageSize?: number;
-  /** Max distinct event IDs remembered for cross-poll de-duplication. Defaults to 10,000. */
+  /**
+   * Size of the LRU window of recent event IDs used for cross-poll
+   * de-duplication. Within this window a repeated event ID (e.g. the same page
+   * re-returned under retry) is suppressed, so downstream consumers see each
+   * event exactly once. Defaults to 1024.
+   */
   dedupCacheSize?: number;
   /** Interval for the self-driving {@link SorobanSubscriber.start} poll loop. Defaults to 2000ms. */
   pollIntervalMs?: number;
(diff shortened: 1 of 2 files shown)


determined-001:
Excellent work @chizzy0011! Replacing the FIFO `Set` with an LRU window for event ID de-duplication is the right approach — it bounds memory under high-throughput scenarios and correctly suppresses repeated events under retry without growing unboundedly. The test coverage for the eviction behavior is thorough. Closes #635.

Issue #632, opened 2026-06-26
1.65 — engine.unsubscribeContract and unsubscribeAllContracts


determined-001:
**Complexity:** Trivial — 100 Points · `type:feature`

**Milestone:** M1 — Soroban event subscription

**Context**
`engine.unsubscribe(address)` exists for classic; contract subscriptions need the inverse.

**Problem**
No engine-level entry point to remove contract subscriptions.

**What needs to be done**
1. Add `engine.unsubscribeContract(config)` — stops the matching watcher, removes registry entry.
2. Add `engine.unsubscribeAllContracts()` — stops every contract watcher without halting Soroban subscriber.

**Key files**
- `packages/pulse-core/src/EventEngine.ts`
- New: `packages/pulse-core/test/EventEngine.unsubscribeContract.test.ts`

**Done when**
- `unsubscribeContract` stops the matching watcher. `unsubscribeAllContracts` leaves Horizon stream running.

chizzy0011:
@chizzy0011 has applied to work on this issue as part of the **Stellar Wave Program**'s 6th wave.



> Good evening maintainer, I want to claim this issue because implementing explicit cleanup routines and subscription lifecycle controls matches my background in state management and reactive event systems. Providing developers with clean entry points to tear down active listeners is essential for preventing memory leaks and managing system resources efficiently. I have strong experience writing resource management layers and lifecycle utilities for asynchronous event structures. I know how to maintain robust internal registries where active watchers can be uniquely identified, isolated, and safely terminated without disturbing adjacent processes or broader long-lived network connections. My background includes building comprehensive batch cleanup mechanisms that flush active states while protecting critical infrastructure routes. I am highly comfortable writing targeted eviction routines that loop through contract subscription pools, stop each running watcher cleanly, and wipe the associated registry fields entirely while keeping the underlying Horizon streams completely operational. For the implementation, I focus on delivering clean code additions and comprehensive testing. I will ensure that both unsubscribe utilities plug seamlessly into the EventEngine class architecture and that the new test file thoroughly verifies teardown behaviors under various load scenarios. I am ready to review your current EventEngine state registry configurations and begin implementing the unsubscribe methods immediately.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/7df66d15-13fc-4dcc-94a7-2dc9a03d6aa1) or assign @chizzy0011 to this issue.

PR #793, opened 2026-06-28
feat implemented engine unsubscribeContract

### LINKED ISSUE #306: 1.65 — engine.unsubscribeContract

**Context**
`engine.unsubscribe(address)` exists for classic; contract subscriptions need the inverse.

**Problem**
The only way to drop a contract subscription is `watcher.stop()`; there is no engine-level entry point.

**What needs to be done**
1. Add `engine.unsubscribeContract(config)` that stops the matching watcher.
2. Document that the engine's internal contract-subscription registry is keyed by the same canonicalised key used for dedup in 1.48.

**Key files**
- `packages/pulse-core/test/EventEngine.unsubscribeContract.test.ts`.

**Done when**
- `engine.unsubscribeContract(config)` stops the matching watcher and removes the registry entry.

firstJOASH:
closes #306


firstJOASH:
please merge PR

determined-001 (changes requested):
Thanks for the PR, @firstJOASH. The changes here (removing two periods from `CODE_OF_CONDUCT.md`) don't implement `engine.unsubscribeContract` as requested by issue #306.

To close #306, the implementation should be in `packages/pulse-core/src/EventEngine.ts` (adding an `unsubscribeContract` method) along with a test in `packages/pulse-core/test/`. Please update this PR with the actual implementation.

firstJOASH:
i’ll resolve that 
accidentally pushed the original changes to the wrong repo
i’ll update the PR as soon as possible

determined-001 (changes requested):
Thanks for iterating, @firstJOASH, but this PR can't move forward as-is:

1. **Feature already on main** — `engine.unsubscribeContract` (both the `string` and `ContractSubscriptionConfig` overloads) was already implemented and merged via PR #632. It's live in `packages/pulse-core/src/EventEngine.ts` on `main` today, so issue #306 is already resolved. Your diff only adds a 6-line no-op test plus a JSDoc line on top of the existing implementation.
2. **Out-of-scope file** — `CODE_OF_CONDUCT.md` is still in the diff (removing two periods); that's an unrelated docs change.

Since #306 is already closed out by #632, I'd recommend closing this PR. If you found a genuine gap in the existing `unsubscribeContract` (a missing case, a bug), please open an issue describing it and we can take a targeted fix.

PR #795, opened 2026-06-28
fix(security): enforce maxBodyBytes limit in webhook verifiers

### LINKED ISSUE #392: 2.55 — Bound the verifier's input size
**Complexity:** Medium — 150 Points · `type:security`

**Milestone:** v1.0 hardening

**Context**
[`SECURITY.md`](../SECURITY.md#pulse-webhooks) recommends a 100kb body cap. The verifier does not enforce.

**Problem**
A malicious sender can ship a multi-megabyte payload and exhaust the receiver's memory before signature verification fails.

**What needs to be done**
1. Add `maxBodyBytes` parameter to both verifiers (default 100_000).
2. Reject longer bodies before HMAC computation.
3. Document the parameter.

**Key files**
- `packages/pulse-webhooks/src/edge.ts`.

**Done when**
- A payload exceeding `maxBodyBytes` is rejected with a distinct error before HMAC runs.

Joseph-1-Duro:
## Linked issue

Closes #392

## Test plan

## Breaking changes

None / Yes (describe below)


determined-001 (changes requested):
Thanks for the PR, @Joseph-1-Duro. Unfortunately the `maxBodyBytes` enforcement for issue #392 already shipped to `main` via PR #733 — it's present in both verifiers today:

- `packages/pulse-webhooks/src/edge.ts` — `const maxBodyBytes = options.maxBodyBytes ?? 100_000;` with the oversize-payload rejection.
- `packages/pulse-webhooks/src/index.ts` — the same guard on the Node verifier.

So issue #392 is already resolved and this PR duplicates it. I'd recommend closing it. If your implementation covers a case the merged version misses (a different verifier entrypoint, a streaming path, etc.), let me know specifically what's different and we can look at landing just that delta.

PR #805, opened 2026-06-29
feat(pulse-core): emit event.decode_failed watcher notification

### LINKED ISSUE #361: 1.120 — EventEngine.normalize returns decode-error events

**Milestone:** M2 — ABI Registry

**Context**
Today a decode failure is silently dropped from `decodedData` (1.117).

**Problem**
Operators have no signal that decoding is failing for a contract.

**What needs to be done**
1. Emit an `event.decode_failed` watcher notification with `{ contractId, eventId, error }`.
2. Document it in the watcher events table.

**Key files**

**Done when**
- A misregistered contract triggers `event.decode_failed`; the event itself is still delivered with `decodedData = undefined`.

Meet-hybrid:
Closes #361

### DIFF
--- a/packages/pulse-core/src/EventEngine.ts
+++ b/packages/pulse-core/src/EventEngine.ts
@@ -114,6 +114,26 @@ const STELLAR_MAX_TRUSTLINE_LIMIT = "922337203685.4775807";
 
 const noop: Logger = { info: () => {}, warn: () => {}, error: () => {} };
 
+const KNOWN_OPERATION_TYPES = new Set([
+  "payment",
+  "set_options",
+  "create_account",
+  "manage_sell_offer",
+  "manage_buy_offer",
+  "bump_sequence",
+  "manage_data",
+  "change_trust",
+  "account_merge",
+  "create_claimable_balance",
+  "claim_claimable_balance",
+  "liquidity_pool_deposit",
+  "liquidity_pool_withdraw",
+  "allow_trust",
+  "set_trust_line_flags",
+  "contract_invocation",
+  "contract_event",
+]);
+
 /**
  * Produces a stable, order-independent string key for a ContractFilter array.
  * Used to deduplicate subscribeContract(config) calls.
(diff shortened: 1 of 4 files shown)


determined-001 (changes requested):
Thanks for the clean re-submission, @Meet-hybrid! The implementation direction looks right for issue #361. However, `typecheck-packages` is failing on both Node 20 and 22 (main typechecks cleanly, so this is PR-specific).

Please run `pnpm -r typecheck` locally, fix the type errors, and push the fix.

determined-001 (changes requested):
Thanks for the lint fix, @Meet-hybrid! Two issues remain:

1. **typecheck-packages still failing** on both Node 20 and Node 22. Please run `pnpm -r typecheck` locally and fix the type errors.

2. **Out-of-scope file added** — your latest commit introduces `packages/pulse-webhooks/src/index.ts` with changes to `doAttempt` error handling. This is unrelated to issue #361 (event.decode_failed watcher notification in pulse-core). Please revert the pulse-webhooks change and keep only the pulse-core changes for this PR.

determined-001 (changes requested):
Closing this PR. A few things:

1. **Issue #361 is already solved on `main`.** It emits `event.decode_failed` with `{ contractId, eventId, error }` when a Soroban contract event's ABI spec lookup fails (`EventEngine.ts` `emitDecodeFailedNotification`, wired into the `contract.emitted` enrichment path). That matches #361's "Done when" criteria exactly.

2. **This PR targets a different, genuinely real gap** — worth acknowledging: `_normalize()`'s fallthrough (an unrecognized top-level Horizon operation `type`) returns `null` silently, with no log or notification at all. That's a real blind spot the existing `event.decode_failed` mechanism doesn't cover. However, this PR reuses the same `"event.decode_failed"` event name for that different failure mode with a different payload shape (`operationType`/`record` instead of `contractId`/`eventId`/`error`), which would make the same event carry two incompatible shapes depending on which code path fired it. I've filed a new issue to track this properly, with a distinct event name.

3. **The second commit duplicates already-merged work** — the `doAttempt` undefined-return fix was already shipped in #808.

4. **This branch is severely out of date.** Diffed against current `main` it shows ~5,950 deletions, reverting large amounts of already-merged work (`spec.ts`, `IRegistryStore.ts`, `FileRegistryStore.ts`, several doc pages, etc.). Nothing to salvage by rebasing.

**Going forward:** please branch off current `main` rather than an old fork snapshot, and check whether an issue is already resolved before opening a PR against it.

PR #938, opened 2026-07-27
feat: add deployment provenance verification to CI

### LINKED ISSUE #894: 8.7 Deployment provenance: verify committed IDs match built WASM
**Labels:** `type:security` `area:contracts`
**Effort:** Medium
**Depends on:** 8.1

#### Description

`contracts/deployed.testnet.json` becomes the trust anchor for the whole
registry: consumers resolve specs through the contract ID it names.

Nothing currently prevents that file from drifting from the source in
`contracts/registry`. A contract deployed from an unreviewed local build would
be indistinguishable from one built off `main`.

#### Acceptance criteria

- [ ] CI builds the contracts reproducibly using the pinned toolchain and compares the WASM hash to `wasmHash` in `contracts/deployed.testnet.json`
- [ ] A mismatch fails the job with a diff of expected vs actual
- [ ] The deployed WASM hash is fetched from the network and asserted equal to the locally built hash
- [ ] Changing contract source without redeploying turns the job red
- [ ] `contracts/README.md` documents the reproducible-build procedure

#### Implementation notes

1. Add the job to `.github/workflows/contracts.yml`.
2. Pin the toolchain from `contracts/rust-toolchain.toml` so the build is
   byte-reproducible across runners.

#### Affected files

- `.github/workflows/contracts.yml`
- `contracts/deployed.testnet.json`
- `contracts/rust-toolchain.toml`
- `contracts/README.md`

Stephan-Thomas:
## Linked issue

Closes #894

## What changed and why

This PR adds deployment provenance verification to the contracts CI pipeline to ensure the deployed registry contract is reproducibly built from the committed source. The workflow now builds the contracts using the pinned Rust toolchain, compares the locally generated WASM hash against the committed `wasmHash` in `contracts/deployed.testnet.json`, and verifies that the deployed WASM retrieved from the testnet matches the local build. Any mismatch causes the workflow to fail with a clear diff, preventing contract metadata from drifting away from the reviewed source. The reproducible build process has also been documented in `contracts/README.md`.

## Test plan

* [ ] New tests added for new behaviour
* [x] Verified the contracts workflow builds using the pinned toolchain from `contracts/rust-toolchain.toml`.
* [x] Confirmed the locally built WASM hash is compared against `contracts/deployed.testnet.json`.
* [x] Confirmed the workflow fetches the deployed WASM hash from testnet and verifies it matches the local build.
* [x] Verified the workflow fails with an expected vs. actual hash diff when hashes do not match.
* [x] Confirmed modifying contract source without updating the deployment metadata causes the provenance check to fail.
* [x] Reviewed `contracts/README.md` to ensure the reproducible build procedure is documented.

## Breaking changes

None.

## Notes for reviewers

The primary review focus should be the reproducibility guarantees in the CI workflow. In particular, verify that the toolchain is fully pinned, the hash comparison uses the same hashing method as `contracts/deployed.testnet.json`, and the network verification step validates the deployed WASM rather than relying solely on the committed metadata. The intent is to make `contracts/deployed.testnet.json` a trustworthy deployment artifact that cannot silently drift from the reviewed source.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -53,3 +53,79 @@ jobs:
       - name: cargo build (release, wasm32v1-none)
         working-directory: contracts
         run: cargo build --release --target wasm32v1-none
+
+  verify-provenance:
+    needs: test
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32v1-none
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: contracts
+
+      - uses: cargo-bins/cargo-binstall@main
+
+      - name: Install stellar-cli
+        run: cargo binstall -y stellar-cli@25.2.0
+
+      - name: cargo build (release, wasm32v1-none)
+        working-directory: contracts
+        run: cargo build --release --target wasm32v1-none
+
+      - name: Verify deployed provenance
+        working-directory: contracts
+        run: |
+          set -euo pipefail
+          
+          if [ ! -f deployed.testnet.json ]; then
+            echo "No deployed.testnet.json found. Skipping provenance verification."
+            exit 0
+          fi
+
+          REGISTRY_WASM="target/wasm32v1-none/release/orbital_abi_registry.wasm"
+          DEMO_WASM="target/wasm32v1-none/release/orbital_demo_emitter.wasm"
+
+          REGISTRY_ID=$(jq -r '.contracts.registry.contractId' deployed.testnet.json)
+          EXPECTED_HASH=$(jq -r '.contracts.registry.wasmHash' deployed.testnet.json)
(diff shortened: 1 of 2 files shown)


determined-001:
Thanks @Stephan-Thomas - #894 asked for the trust anchor to stop being trusted on faith, and this does both halves: local reproducible build compared to `deployed.testnet.json`, then the on-chain WASM fetched from testnet and compared to the local build. The `verify-provenance` job passed against the real deployment, so the recorded hashes are genuinely correct today. Merging - closes #894.

determined-001:
Correction to my merge note above: the `verify-provenance` job did **not** verify real hashes on this run. `contracts/deployed.testnet.json` does not exist on `main`, so the job took its own early-exit branch:

```
if [ ! -f deployed.testnet.json ]; then
  echo "No deployed.testnet.json found. Skipping provenance verification."
  exit 0
fi
```

The job is still correct and still merged - the guard is the right behaviour before deployment. But it is currently a no-op, and it will only start doing real work once the contracts are actually deployed and that file is populated with real contract IDs and `wasmHash` values. Worth knowing so nobody reads a green `verify-provenance` as proof the deployment matches source.

PR #940, opened 2026-07-27
feat: Generate typed event guards from a resolved spec

OLAMIGHTY010:
Closes #904

## Linked issue

Closes #

## Test plan

- [ ] Existing tests pass (`pnpm test`)
- [ ] New tests added for new behaviour
- [ ] Typechecks pass (`pnpm -r typecheck`)
- [ ] Manually tested against testnet / mainnet (describe what you tested)

## Breaking changes

None / Yes (describe below)

Issue #904, opened 2026-07-27
10.2 Generate typed event guards from a resolved spec


determined-001:
**Labels:** `type:feature` `type:dx` `area:abi-registry`
**Milestone:** Wave 2.2 - orbital codegen
**Effort:** Medium
**Depends on:** none

#### Description

`packages/pulse-core/src/eventTypeGuard.ts` provides hand-written narrowing for
the built-in taxonomy; `packages/abi-registry/src/generate.ts` is the codegen
foundation.

Contract events get types but no narrowing helpers, so consumers hand-write
`if (event.topic === ...)` checks the compiler cannot verify - exactly the bug
class the typed-event thesis exists to eliminate.

#### Acceptance criteria

- [ ] Codegen emits an `isXxxEvent(e): e is XxxEvent` guard per event, alongside the type and Zod schema
- [ ] Codegen emits a discriminated union of a contract's events plus an exhaustive `switch` helper
- [ ] Generated output includes a `.test-d.ts` that fails to compile when an event is added to the spec but not handled
- [ ] Each guard validates at runtime through its Zod schema, not by shape-guessing

#### Implementation notes

1. Mirror the existing exhaustiveness pattern in
   `packages/pulse-core/test/types.exhaustive.test-d.ts`.

#### Affected files

- `packages/abi-registry/src/generate.ts`
- `packages/pulse-core/src/eventTypeGuard.ts`

OLAMIGHTY010:
@OLAMIGHTY010 has applied to work on this issue as part of the **Stellar Wave Program**'s 7th wave.



> I would like to work on this issue.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/57dbd2ef-b1b4-4d14-a110-4057be94fd8c) or assign @OLAMIGHTY010 to this issue.

Issue #18, opened 2026-03-26
fix(build): replace build placeholder with real tsc compilation for all publishable packages


determined-001:
## Context

`.github/workflows/build.yml` exists to validate that all packages compile successfully on every push to `main`. Currently the entire job is a single `echo` placeholder — no compilation is ever run. The workflow reports green regardless of the state of the codebase, giving false confidence.

## Problem

A broken build on `main` goes completely undetected. Any contributor merging code that fails `tsc` compilation would not be caught, silently breaking the project for everyone downstream.

## What needs to be done

Replace the placeholder build job with real steps that:

1. Check out the repo and set up Node
2. Install dependencies via `npm ci` at the root
3. Run `tsc` (or the configured build script) for each publishable package: `packages/pulse-core`, `packages/pulse-webhooks`, and `packages/pulse-notify`
4. Run the Next.js build for `apps/web` (`next build` or `npm run build --workspace=apps/web`)
5. Skip `apps/server` — it has no `tsconfig.json` and is not compiled
6. Fail the workflow if any step fails

## Key files

* `.github/workflows/build.yml` — replace the placeholder job with real per-package compilation steps

## Implementation notes

* Run each package as a **separate named step** so failures are easy to pinpoint in the Actions log
* If a `build` script exists in each package's `package.json`, prefer `npm run build --workspace=<pkg>` over calling `tsc` directly
* Do not introduce new build tooling — only wire up what already exists
* `apps/server` should be explicitly skipped with an inline comment explaining why (no `tsconfig.json`, uses `tsx`)

## Done when

* The build workflow runs real compilation — no `echo` placeholder
* A deliberate TypeScript error in any of the three packages causes the build workflow to fail
* `apps/web` Next.js build is included
* `apps/server` is explicitly skipped with a comment
* The workflow passes cleanly on the current `main` branch

cybermax4200:
@cybermax4200 has applied to work on this issue as part of the **Stellar Wave Program**'s 3rd wave.


> Hi 👋, I’d like to take on this issue.
> 
> I’m particularly interested in improving CI reliability and developer confidence, and fixing a placeholder build workflow is critical to ensuring the codebase remains stable.
> 
[...]

PR #959, opened 2026-07-28
feat(pulse-notify): add useContractEvent hook generated from registry schema (#905)

### LINKED ISSUE #905: 10.3 useContractEvent hook generated from the registry schema
**Labels:** `type:feature` `area:pulse-notify`
**Milestone:** Wave 2.2 - orbital codegen
**Effort:** Medium
**Depends on:** none

#### Description

`packages/pulse-notify` ships `useStellarEvent`, `useStellarPayment`,
`useStellarActivity`, `useContractState` and Suspense variants, but no hook
typed by a *specific contract's* schema.

React hooks over registry-typed contract events is the differentiation against
`stellar contract bindings typescript`, which emits raw types and nothing else.

#### Acceptance criteria

- [ ] `useContractEvent<T>({ contractId, event, filter })` returns validated, typed events
- [ ] Codegen emits contract-specific wrappers (e.g. `useSwapExecuted()`) bound to the generated types
- [ ] One subscription exists per `(contractId, event)` regardless of hook-instance count, asserted by test
- [ ] Subscription count returns to zero after unmount, asserted by test
- [ ] A runnable snippet is in `docs/COOKBOOK.md`

#### Implementation notes

1. Reuse `packages/pulse-notify/src/connectionPool.ts` rather than opening a
   socket per hook.

#### Affected files

- `packages/pulse-notify/src/` (new hook)
- `packages/pulse-notify/src/connectionPool.ts`
- `packages/abi-registry/src/generate.ts`
- `docs/COOKBOOK.md`

Biba-tech-hub:
## Overview

This PR implements `useContractEvent<T>` hooks generated from the registry schema — the key differentiation against `stellar contract bindings typescript` which emits raw types and nothing else. Codegen now emits per-event React hooks (e.g. `useSwapExecuted()`) with Zod runtime validation, and the hook itself accepts an optional schema for runtime data validation.

## Related Issue

Closes #905

### Hook generator (abi-registry)

* **[ADD]** `packages/abi-registry/src/generate.ts` — `generateReactHooks()` / `generateContractHooks()` emit per-event wrapper hooks like `useSwapExecuted({ contractId, serverUrl })` that wrap `useContractEvent` with the correct Zod schema pre-applied
* **[MODIFY]** `GeneratedContractArtifacts` now includes an optional `hooks` field
* **[MODIFY]** `packages/abi-registry/src/index.ts` — Exports `generateContractHooks`, `generateContractArtifacts`, `generateContractTypes`

### Hook runtime (pulse-notify)

* **[MODIFY]** `packages/pulse-notify/src/index.ts` — `useContractEvent` now accepts an optional `schema` option for runtime event data validation via `safeParse`. Events whose `data` fails schema validation are filtered out
* **[ADD]** `EventSchema<T>` type exported for consumer use
* **[ADD]** `packages/pulse-notify/test/useContractEvent.test.tsx` — 3 new tests:
  - Schema validation: invalid data is filtered, valid data passes through
  - Connection sharing: one EventSource for 3 hook instances with the same `(contractId, topics)`
  - Cleanup: subscription count returns to zero after all instances unmount

### Documentation

* **[ADD]** `docs/COOKBOOK.md` — New recipe #14 with 3-step walkthrough (generate → use with schema → use auto-generated hooks)

## Verification Results

```
pnpm test (pulse-notify):
✅ useContractEvent › validates event data against optional schema
✅ useContractEvent › shares one connection per (contractId, event) regardless of hook-instance count
✅ useContractEvent › subscription count returns to zero after all instances unmount
✅ all existing tests pass
```

| --- | --- |
| `useContractEvent({ contractId, event, filter })` returns validated, typed events | ✅ schema option validates at runtime |
| Codegen emits contract-specific wrappers bound to generated types | ✅ `generateContractHooks()` emits `useSwapExecuted()` etc. |
| One subscription per (contractId, event) regardless of hook-instance count | ✅ asserted by test |
| Subscription count returns to zero after unmount | ✅ asserted by test |
| Runnable snippet is in docs/COOKBOOK.md | ✅ Recipe #14 |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,8 @@ jobs:
         run: pnpm tsc --noEmit -p packages/pulse-webhooks/tsconfig.json
       - name: Typecheck pulse-notify
         run: pnpm tsc --noEmit -p packages/pulse-notify/tsconfig.json
+      - name: Typecheck anchor-sdk
+        run: pnpm tsc --noEmit -p packages/anchor-sdk/tsconfig.json
       - name: Validate ABI specs
         run: pnpm --filter @orbital-stellar/abi-registry validate
       - name: Test packages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ Per-package changelogs live in each package directory.
   are frozen and moved to an explicit Frozen section in `ROADMAP.md` with
   per-item rationale and an unfreeze procedure. No shipped code is removed -
   planned scope only.
+- Docs aligned with the refocused roadmap: README reframed around the
+  decoding standard; open-source-policy and proposal forward-references
+  updated; frozen-scope references removed from forward-looking docs.
 
 ### Fixed
 
--- a/PROGRESS.md
+++ b/PROGRESS.md
@@ -1,44 +1,17 @@
 # Orbital: Progress & Status Report
 
-**Last Updated:** 2026-07-06
+**Last Updated:** 2026-07-16
 **Project Status:** Core SDK family - Shipped ✅
 
 ---
 
-## Executive Summary
+## Status
 
-Orbital is **Stellar's open-source real-time event SDK family** - four MIT-licensed packages on npm that give any Stellar developer typed event subscriptions (Horizon and Soroban), signed webhook delivery with durable retry, typed ABI decoding, and React hooks, without re-implementing the plumbing.
+Orbital is **the typed event layer for Stellar** - four MIT-licensed packages on npm that give any Stellar developer typed event subscriptions (Horizon and Soroban), signed webhook delivery with durable retry, typed ABI decoding, and React hooks, without re-implementing the plumbing.
 
-**Current Status:** The full classic operation taxonomy, Soroban contract event subscription, cursor persistence, durable webhook retry queues, the ABI registry client, and edge-runtime webhook verification are all shipped and tested. Next up is the Phase 2 SDK family (`@orbital-stellar/hooks`, `@orbital-stellar/payments`, `@orbital-stellar/auth`) - see [`ROADMAP.md`](ROADMAP.md).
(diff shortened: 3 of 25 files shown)


determined-001 (changes requested):
Thanks @Biba-tech-hub - a registry-typed `useContractEvent` is the differentiator #905 exists for, and emitting contract-specific wrappers from codegen is the right call. Two failures.

**1. `typecheck-packages` fails on a real test failure** (both Node 20 and 22):

```
FAIL  test/useContractEvent.test.tsx > useContractEvent Hook > validates event data against optional schema
TestingLibraryElementError: Unable to find an element with the text: true, which matches selector '[data-testid="connected"]'

<body>
  <div>
    <div data-testid="event">null</div>
  </div>
</body>
```

The test harness only renders `data-testid="event"` - there is no `connected` node in the tree at all, so the `waitFor` can never resolve. Either the test component needs to render the connection state, or the assertion should wait on `event` instead.

**2. `format:check` fails.**

```
[warn] packages/abi-registry/src/generate.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

`pnpm format:write` fixes it.

Heads up on a conflict: **#940 merged into `main` and rewrote `packages/abi-registry/src/generate.ts`** to emit typed event guards, a discriminated union and a `.test-d.ts` for #904. Please rebase onto `main` before pushing the fixes - your hook codegen and its guard codegen both live in that file now.

Once it is green, please also confirm the third criterion on #905: **one subscription per `(contractId, event)` regardless of how many hook instances mount.** That is the one that decides whether this scales past a demo page, and I did not see a test pinning it.

Biba-tech-hub:
@determined-001 addressed the review:

1. **Schema validation test** — `SchemaTestComponent` now renders `data-testid="connected"` (same as the other harnesses), so the `waitFor`/`findByText` can resolve.
2. **Rebased onto `main`** — merged hook codegen with #940’s guard/`testDts` codegen in `packages/abi-registry/src/generate.ts` (`GeneratedContractArtifacts` keeps `guards`, `testDts`, and `hooks`).
3. **Format** — `generate.ts` / tests pass `prettier --check`.
4. **Connection pooling criterion** — covered by `shares one connection per (contractId, event) regardless of hook-instance count` (asserts `MockEventSource.instances.length === 1` and pool size `1` for three mounted hooks) plus the unmount cleanup test. All 75 pulse-notify tests pass locally.

Cookbook recipe is numbered **#18** after rebasing onto main’s expanded TOC.

Ready for another look.

determined-001:
Merged — thanks, the optional schema on `useContractEvent` plus the generated per-event wrappers is exactly what 10.3 was after.

One maintainer commit before merge (`fix(pulse-notify): read useContractEvent schema through a ref`): the subscription effect does not list `schema` in its dependency array — correctly, since resubscribing whenever a consumer passes an inline schema object would tear down the stream on every render — but it read `schema` directly from the closure, so whichever schema was present on first render was pinned forever. It now goes through a ref like `filter` and `onEvent` do.

The only red check was Vercel ("Authorization required to deploy"), which fails on every fork PR here and is unrelated.

PR #937, opened 2026-07-27
feat(abi-registry): consume SEP-48 embedded event specs natively (#903)

OLAMIGHTY010:
- Add SpecSource type (sep48|registry|wellKnown|discovery) and ResolvedSpec wrapper to spec.ts

- Add Sep48EmbeddedClient in discovery/ that fetches WASM and returns specs only when embedded #[contractevent] entries are present

- Add getResolvedSpec() to ChainedAbiRegistryClient with provenance tracking and conflict warning (emits console.warn with both hashes when specs disagree)

- Wire Sep48EmbeddedClient as first link in createDefaultAbiRegistryClient precedence chain

- Extend ChainedAbiRegistryClient tests with provenance, precedence, and conflict warning coverage

- Add Sep48EmbeddedClient tests with real demo-emitter.wasm fixture and non-synthetic precedence integration test

All 223 tests pass (18 test files). Issue 7.9 precedence tests remain green.

Closes #903

## Linked issue

Closes #

## Test plan

- [ ] Existing tests pass (`pnpm test`)
- [ ] New tests added for new behaviour
- [ ] Typechecks pass (`pnpm -r typecheck`)
- [ ] Manually tested against testnet / mainnet (describe what you tested)

## Breaking changes

None / Yes (describe below)

Issue #915, opened 2026-07-27
12.1 Hosted read API for schemas, taxonomy and labels


determined-001:
**Labels:** `type:feature` `area:apps-web`
**Milestone:** Wave 2.4 - Hosted registry
**Effort:** High
**Depends on:** 8.3, 11.1, 11.2

#### Description

Resolving a spec today means an RPC round trip per unknown contract. At any real
request rate that is slow and rate-limited, and every integration is coupled to
one RPC endpoint's availability.

#### Acceptance criteria

- [ ] `GET /v1/registry/spec/:contractId[?version=]`, `/v1/registry/taxonomy`, `/v1/registry/labels` and `/v1/registry/health` all serve live data
- [ ] Every response includes `servedFrom`, `asOfLedger` and the spec hash
- [ ] Reads go through a read-through cache with an explicit TTL and `stale-while-revalidate`; stale data is never served silently
- [ ] `/health` reports the last-sync ledger
- [ ] An OpenAPI description is published at `/v1/registry/openapi.json`

#### Implementation notes

1. Version the path from day one - the client in 12.2 pins to `/v1/`.

#### Affected files

- `apps/web/app/api/registry/` (new)
- `packages/abi-registry/src/OnChainAbiRegistryClient.ts`

khanavi272-spec:
@khanavi272-spec has applied to work on this issue as part of the **Stellar Wave Program**'s 7th wave.



> I would like to work on this issue 


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/0ce2959e-2828-4e01-900a-090eb47b6f86) or assign @khanavi272-spec to this issue.

PR #965, opened 2026-07-28
feat:  Hosted-API client with automatic chain fallback

### LINKED ISSUE #916: 12.2 Hosted-API client with automatic chain fallback
**Labels:** `type:feature` `area:abi-registry`
**Milestone:** Wave 2.4 - Hosted registry
**Effort:** Medium
**Depends on:** 12.1

#### Description

`ChainedAbiRegistryClient` already composes resolution sources behind an LRU/TTL
cache. If the hosted API becomes a link in that chain, an outage there must
degrade to a direct chain read - otherwise the operated service becomes a single
point of failure for an MIT library.

#### Acceptance criteria

- [ ] `HostedAbiRegistryClient` hits the `/v1/` endpoints with a configurable base URL and timeout
- [ ] It sits ahead of the on-chain client in the default chain and falls through on timeout, 5xx, or hash mismatch
- [ ] Returned spec hashes are verified against chain on a sampled basis (configurable, default 1 in 20), with a hard error logged on mismatch
- [ ] One config flag opts out of the hosted link entirely for chain-only resolution
- [ ] Tests cover hosted hit, timeout fallback, and hash-mismatch fallback with error log

#### Implementation notes

1. A hash mismatch means the hosted service returned a spec the chain does not
   attest to - fail loudly, never prefer the faster answer.

#### Affected files

- `packages/abi-registry/src/HostedAbiRegistryClient.ts` (new)
- `packages/abi-registry/src/ChainedAbiRegistryClient.ts`
- `packages/abi-registry/src/createDefaultAbiRegistryClient.ts`

olawale880:
## Linked issue

Closes #
close #916

## Test plan

## Breaking changes

None / Yes (describe below)


determined-001 (changes requested):
Thanks @olawale880 - #916 matters precisely because the hosted API must not become a single point of failure for an MIT library, and putting `HostedAbiRegistryClient` ahead of the on-chain client with fall-through is the right shape. Two fixes, one design question.

**1. `typecheck-packages` fails - and the failing test is itself wrong.**

```
FAIL  test/HostedAbiRegistryClient.test.ts > URL construction > strips a trailing slash from the base URL
AssertionError: expected 'https://abi.orbital.dev/v1/specs/CCW6…' not to match /\/\//
```

The assertion is `expect(url).not.toMatch(/\/\//)`, but every absolute URL contains `//` right after the scheme, so it can never pass. Assert on the path instead - e.g. check the URL does not contain `/v1//`, or compare against the exact expected string.

**2. `format:check` fails.**

```
[warn] packages/abi-registry/src/HostedAbiRegistryClient.ts
[warn] packages/abi-registry/test/HostedAbiRegistryClient.test.ts
```

`pnpm format:all`.

**3. The default base URL points at a host that appears nowhere else in this repo.**

`https://abi.orbital.dev` - there is no other reference to that domain in the codebase, and #915 (the hosted API this client talks to, PR #964) is still open and unmerged, so the `/v1/` endpoints do not exist yet. Please confirm the host, and put it in `registryConstants.ts` as a single named constant rather than inline. A default that silently fails to resolve is worse than no default - and since this client sits **ahead** of the on-chain one in the chain, it must degrade cleanly and quickly rather than adding a timeout to every resolution.

Then, the criterion that decides whether this is safe to put first in the chain:

> Returned spec hashes are verified against chain on a sampled basis (configurable, default 1 in 20), with a hard error logged on mismatch

I could not find the sampled verification in the diff. Without it the hosted API can serve a wrong spec indefinitely and nothing notices - which is the whole risk the issue is guarding against. Please add it, with a test covering the mismatch path.

Note #937 has landed on `main` and reworked `createDefaultAbiRegistryClient` to insert `Sep48EmbeddedClient` first in the chain, so please rebase and decide explicitly where the hosted client sits relative to the embedded SEP-48 source.

determined-001 (approved):
Approved. This is the strongest contributor PR in the current queue — all five of #916's acceptance criteria are genuinely met, not approximated:

- `/v1/specs/{contractId}` and `/v1/specs/{contractId}?ledger={n}` with configurable `baseUrl` and `timeoutMs`
- Sits ahead of the on-chain client in the default chain, falling through on timeout, 5xx, and hash mismatch
- Sampled hash verification, configurable via `sampleRate`, defaulting to 20 (1 in 20), with `0` to disable
- `chainOnly` opts out of the hosted link entirely
- Tests cover hosted hit, timeout fallback, and hash-mismatch fallback with the error log — plus 404, malformed JSON, DNS failure, trailing-slash handling, contract-ID encoding, and separate cache keys for `getSpec` / `getSpecAt`

The detail I want to single out is that a hash mismatch logs an error and falls through, while timeouts and 5xx fall through **silently**. That distinction is exactly right and is the thing #916's implementation note was warning about — a mismatch means the hosted service returned a spec the chain does not attest to, which is a correctness problem worth shouting about, whereas an outage is just an outage. You also got the caching right by not caching a mismatch.

I fixed four things on your branch rather than sending it back.

**1. The branch was 17 commits behind `main`** and predated three changes to `createDefaultAbiRegistryClient`: the `Sep48EmbeddedClient` link, the `readEnv()` environment indirection, and the `ChainedAbiRegistryClient` return type. I merged `main` in (not a rebase — I did not want to rewrite your history) and resolved the conflict by combining both option surfaces:

```ts
export type CreateDefaultAbiRegistryClientOptions = {
  rpcUrl?: string;
  chainOnly?: boolean;
};
```

Resulting precedence, which preserves what #916 asks for while keeping what landed on `main`:

1. SEP-48 embedded (when `rpcUrl` given)
2. Bundled well-known
3. **Hosted** (skipped when `chainOnly` or no base URL)
4. On-chain

I also routed the hosted base URL through `readEnv("ORBITAL_HOSTED_REGISTRY_BASE_URL")` so it behaves like the other registry constants rather than being the one that ignores the environment.

**2. Two test bugs that would never have passed.**

```ts
expect(url).not.toMatch(/\/\//);
```

`https://` contains `//`, so this assertion could not pass for any absolute URL — it was failing on a correct implementation. Your trailing-slash stripping works; the check was looking in the wrong place. Now:

```ts
expect(new URL(url).pathname).not.toMatch(/\/\//);
```

And on the next line, `expect(url).toStartWith(...)` — `toStartWith` is not a Vitest matcher, so this would have thrown `is not a function` once the previous line stopped failing first. Replaced with `expect(url.startsWith(...)).toBe(true)`.

**3. A redundant `eslint-disable no-console`** on the mismatch log, which lint flagged as an unused directive. `console.error` is permitted there, so the directive was the only problem.

**4. Prettier** on both files for `format:check`.

Separately, `lint` was failing on this PR for reasons unrelated to it — six pre-existing warnings on `main` that tripped `--max-warnings 0`. Fixed on `main` in `e738c56` and merged in here.

Everything is green now. One optional follow-up, not a blocker: `shouldSample()` uses `Math.random()`, so sampling is nondeterministic in production, which is correct, but it means a mismatch may go undetected for a while on a low-traffic contract. If you want to tighten that later, sampling the first request per contract-ID and then falling back to the rate would guarantee at least one verification per contract per process. Worth its own issue rather than expanding this PR.

Nice work — thanks for the thorough test coverage, and sorry the stale base and the `main`-side lint problem made this harder than it should have been.

PR #979, opened 2026-08-02
feat(examples): runnable express-starter backend reference

### LINKED ISSUE #897: 9.2 orbital-express-starter boilerplate

determined-001:
## Linked issue

Closes #897 (9.2 orbital-express-starter boilerplate)

## What changed and why

`docs/COOKBOOK.md` describes the ingest → cursor → signed-webhook composition in prose. Prose does not compile. This is the version that runs.

`examples/express-starter/` — an EventEngine watching configured addresses, a `WebhookDelivery` per address, a cursor that survives restarts, and an Express receiver that verifies what arrives. The process is both sender and receiver by default, so `pnpm dev` shows the entire loop locally with no other services.

### Express, and what it cost

The issue asks for a justification. **Express**, because this is a *reference*: a reader should recognise the shape and lift it into the app they already run, and Express is still what most Node backends use. The composition itself is framework-agnostic — `createReceiver()` is 60 lines and the only Express surface is `app.post`.

It did surface one thing worth knowing: **Express 5 requires `path-to-regexp@8`, and this workspace pinned `path-to-regexp: 0.1.13` globally** for [GHSA-9wv6-86v2-598j](https://github.com/advisories/GHSA-9wv6-86v2-598j). The unscoped override dragged Express 5's router back to a version whose API it cannot use — `TypeError: pathRegexp.match is not a function` on every route. The override is now `"path-to-regexp@0": 0.1.13`, so the advisory stays covered for anything on the 0.x line and Express 5 resolves 8.4.2. Same pattern the `js-yaml@3` pin already uses.

### The three guarantees, each with a test

**Resume after restart.** The cursor is `PostgresCursorStore` when `DATABASE_URL` is set, `FileCursorStore` otherwise — the file store is not a toy, it survives a restart, which is what the guarantee needs. Tests assert a second store instance over the same directory reads what the first persisted, that an unseen stream reports `null` rather than a stale cursor, and that it advances monotonically.

**Signature verification.** Two details the receiver gets right on purpose: it verifies against the **raw body** (`express.json()` re-serialises, and the bytes are not always identical to what was signed), and it **rejects rather than logs** — a tampered payload gets a 401 and never reaches application code. Tests cover a valid delivery, a payload with one digit changed after signing, a wrong secret, and missing headers.

**Graceful shutdown.** SIGTERM/SIGINT call `engine.stop()` and **await it** before `process.exit`. The test asserts that ordering — `["stop", "exit:0"]` — not merely that a handler is registered, because exiting on the signal without waiting is exactly what loses the final cursor write.

### Also

- `docker-compose.yml` for Postgres, with the `cursor_store` table created on startup
- `.env.example` and a config loader that **refuses to start** on a missing address list, a secret under 16 characters, or a non-numeric port
- README with the step-by-step resume-after-restart walkthrough the issue asks for

## Test plan

- [x] Existing tests pass (`pnpm -r test`, all seven workspaces green)
- [x] New tests added — 14, no containers required
- [x] Typechecks pass
- [x] Coverage gate green
- [ ] Manually tested against testnet — see below

## Not verified here

- **`docker compose up && pnpm start` against real Postgres.** The compose file and the `PostgresCursorStore` wiring are written but I cannot run Docker in this environment, so the Postgres path is untested. The file-cursor path — same guarantee, same code path through `EventEngine` — is covered.
- **A live testnet event end to end.** The receiver is tested against signed payloads directly rather than against Horizon.

Both are worth ten minutes of manual confirmation before this is held up as *the* backend reference.

## Breaking changes

None. New example package; the `path-to-regexp` override change is a narrowing that leaves the advisory covered.

### DIFF
--- a/examples/express-starter/.env.example
+++ b/examples/express-starter/.env.example
@@ -1,0 +1,17 @@
+# Accounts to watch (comma-separated).
+STELLAR_ADDRESSES=GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN
+
+# testnet (default) or mainnet.
+STELLAR_NETWORK=testnet
+
+# Shared HMAC secret. Generate with: openssl rand -hex 32
+WEBHOOK_SECRET=replace-me-with-32-bytes-of-hex
+
+# Where signed deliveries go. Defaults to this process's own receiver.
+# WEBHOOK_URL=http://127.0.0.1:3000/hooks/stellar
+
+# Port for the receiver and /healthz.
+PORT=3000
+
+# Set to use Postgres for the cursor; omit for a file-backed cursor.
+# DATABASE_URL=postgres://orbital:orbital@127.0.0.1:5432/orbital
--- a/examples/express-starter/README.md
+++ b/examples/express-starter/README.md
@@ -1,0 +1,127 @@
+# orbital-express-starter
+
+A runnable backend reference: ingest Stellar events, persist a cursor, deliver
+HMAC-signed webhooks, and verify them on the receiving end. `docs/COOKBOOK.md`
+describes this composition in prose — this is the version that compiles.
+
+Implements issue 9.2 (#897).
+
+## Why Express
+
+The issue allows Express or Fastify. Express, because this is a *reference*: the
+point is that a reader recognises the shape and can lift it into the app they
+already run, and Express is still what most Node backends use. The composition
+is framework-agnostic — `createReceiver()` is 60 lines and the only Express
+surface is `app.post`.
+
+One caveat it exposed: Express 5 needs `path-to-regexp@8`, and this workspace
+pinned `path-to-regexp: 0.1.13` globally for [GHSA-9wv6-86v2-598j](https://github.com/advisories/GHSA-9wv6-86v2-598j).
+The override is now scoped to `path-to-regexp@0`, so the advisory stays covered
+for anything on the 0.x line and Express 5 gets the version its router needs.
+
+## Quickstart
+
+```bash
+cp .env.example .env      # set STELLAR_ADDRESSES and WEBHOOK_SECRET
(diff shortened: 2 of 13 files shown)


determined-001:
Note on the CodeQL round-trip: the receiver originally had a hand-rolled fixed-window limiter, tested and working, but `js/missing-rate-limiting` stayed high because CodeQL recognises rate limiting by library rather than by behaviour. Swapped to `express-rate-limit`, which is the more copyable choice for a reference backend anyway. Behaviour is identical — per-IP, applied before the body parser and before any HMAC work, same 429 envelope and `onRejected` callback. All checks green now.

